PR #1, opened 2026-03-20
125 stabilize JSCompTest

### LINKED ISSUE #125: Make our Jensen-Shannon-Compatibility Test more efficient
***In GitLab by @asishallab on Feb 14, 2026, 05:45 GMT+1:***

I propose the following update.

# Set-Based Jensen–Shannon Compatibility Test

## Definitions

Let  

- $D = \{S_1, \dots, S_K\}$ denote a set of $K$ studies sampling the same nominal biological condition i.e. tissue.
- $G$ denote the set of genes.
- $x^{(S_k)}_{g,i}$ denote the expression value of gene $g \in G$ in replicate $i$ of study $S_k$.
- $n_{S_k,g}$ denote the number of biological replicates available for gene $g$ in study $S_k$.

Per-gene mean expression:

$$
\bar{x}^{(S_k)}_g = \frac{1}{n_{S_k,g}} \sum_{i=1}^{n_{S_k,g}} x^{(S_k)}_{g,i}
$$

Signed residuals:

$$
e^{(S_k)}_{g,i} = x^{(S_k)}_{g,i} -\bar{x}^{(S_k)}_g
$$

Residuals represent replicate-level variability around the gene mean.

---

## Step 1 — Precompute and Sort Mean Expression Values

For each study $S_k$:

- Compute all gene means $\bar{x}^{(S_k)}_g$.
- Sort the vector of gene means once and store sorted indices.

This sorting step is performed once per study and reused for all comparisons.

---

## Step 2 — Construct Shared Mean-Expression Reference Grid

- Pool all gene means across studies:

$$
\{\bar{x}^{(S_k)}_g\}_{k=1}^K
$$

- Merge the sorted arrays (linear-time merge). Important: Implement a merge function for pre-sorted arrays as a helper and use here. Store merged sorted array for later use.
- Define $r$ reference points:

$$
x^\star_1, \dots, x^\star_r
$$

chosen as empirical quantiles of the pooled mean-expression distribution.

- For each reference point $x^\star_j$, define a k-nearest-neighbor (kNN) neighborhood in mean-expression space.

This yields aligned conditioning neighborhoods across all studies.

---

## Step 3 — Construct Conditional Residual Distributions

For each study $S_k$ and reference point $x^\star_j$:

- Collect residuals $e^{(S_k)}_{g,i}$ from genes in the corresponding kNN neighborhood.
- Define a shared residual range $[-R_{D,j}, R_{D,j}]$, where

$$
R_{D,j} = \max_{k=1,\dots,K} R^{(S_k)}_j
$$

and $R^{(S_k)}_j$ is a high quantile (e.g. 95%) of absolute residuals in that neighborhood.

- Partition the range into $M$ fixed bins shared across all studies.
- Construct histograms and normalize to obtain probability mass functions (PMFs):

$$
P_k(e \mid x^\star_j)
$$

---

## Step 4 — Compute Consensus Distribution

For each reference point $x^\star_j$, compute the arithmetic mean distribution:

$$
M_j(e) = \frac{1}{K} \sum_{k=1}^K P_k(e \mid x^\star_j)
$$

This is the standard multi-distribution Jensen–Shannon formulation (no leave-one-out).

---

## Step 5 — Compute Divergence per Study

For each study $S_k$ and reference point $x^\star_j$:

$$
\mathrm{JSD}_k(x^\star_j) = \mathrm{KL} \left( P_k(\cdot \mid x^\star_j) \parallel M_j \right)
$$

Aggregate across reference points using weights $w_j$ proportional to the number of residual samples contributing to each neighborhood:

$$
\mathrm{JSD}_k^{\mathrm{global}} = \sum_{j=1}^r w_j \, \mathrm{JSD}_k(x^\star_j)
$$

---

## Interpretation

- Small $\mathrm{JSD}_k^{\mathrm{global}}$: study follows the consensus residual structure.
- Large $\mathrm{JSD}_k^{\mathrm{global}}$: study deviates in replicate variability, indicating a potential batch effect or biological mismatch.

This procedure naturally generalizes to two, three, or many studies. The computationally expensive steps (sorting and neighborhood construction) are performed once per study, making the method scalable to large multi-study comparisons.

**Assignees:** @qwerdenkerXD, @schroederaaron, @jitu2d, @VivianBass

*Migrated from GitLab: https://gitlab.rlp.net/a.hallab/tensor-omics/-/work_items/124*

### LINKED ISSUE #126: Stabilizing and Optimizing the JSD-Compatibility-Test (JSCompTest)
***In GitLab by @qwerdenkerXD on Feb 19, 2026, 05:06 GMT+1:***

### Motivation

The JSCompTest depends on two interacting resolution parameters:

- $r$: number of mean-expression anchor points $\bar{x}^*_j$
- $k_x$: neighborhood size (kNN) in mean-expression space

Both parameters influence numerical stability, sensitivity, and runtime.
To avoid arbitrary parameter choices and ensure reproducible effect sizes, we introduce a **data-informed parameter search with early stopping based on confidence interval (CI) stability**, followed by permutation-based significance testing only once a stable parameter regime has been identified.

---

### 1. Data-informed grid construction for $(r, k_x)$

We construct a candidate grid of parameter pairs $(r, k_x)$ based on dataset size:

- $r$ is chosen as outlined below
- $k_x$ is chosen relative to the number of genes $G$
  (e.g. $k_x \in \{\lfloor G/(4r)\rfloor, \lfloor G/(2r)\rfloor\}$); **use log-spacing as in the below $r$ grid explanation, please**

This ensures that neighborhoods are sufficiently populated while covering the dynamic range of mean expression.

#### grid for $r$ parameter values

Choosing the anchor grid size r

Let N be the number of genes (mean-expression values) pooled across all studies.

Initial anchor count

Define a data-informed default that grows sublinearly with N:

r0 = clip(ceil(10 * sqrt(N)), r_min, r_max)

Typical values:
r_min = 200
r_max = 1500

Lower bound

Define the minimum resolution to explore:

r_low = max(r_min, ceil(0.2 * r0))

This ensures sufficient coverage while avoiding overly coarse grids.

Candidate sequence (log-spaced)

Generate anchor counts in descending order:

r_l = round(r0 * gamma^l), l = 0, 1, …, L−1

with:
gamma ≈ 0.8 (range 0.75–0.85)
L ≈ 6–8

Clamp values to:
r_low ≤ r_l ≤ r0

Remove duplicates and stop once r_low is reached.

Rationale

Log-spaced reduction probes resolution at constant relative change, which is well suited for detecting stability plateaus. Linear spacing would oversample near r0 and undersample the low-resolution regime.

Example (TCGA-scale)

N = 90,000
r0 ≈ 3000 → clipped to 1500

Resulting sequence (gamma = 0.8):
1500, 1200, 960, 768, 614, 491, 393, 314

This sequence is evaluated in descending order until numerical stability is reached, before running permutation tests.

---

### 2. Enforcing minimum neighborhood overlap, min residual numbers and min per bin numbers

For a parameter pair $(r, k_x)$ to be admissible, **adjacent neighborhoods must overlap** by at least a minimum fraction $\mathcal{o}_x$ (default: $10\%$).

Also enforce that neighborhoods have at least 10 residuals per histogram bin in the mean study.

Pairs violating this constraint are removed.

If too few admissible pairs remain:
- allow $k_x$ to exceed its nominal maximum
- or reduce $r$ adaptively

This guarantees smooth conditioning in mean-expression space.

---

### 3. Ordering of parameter pairs

Remaining parameter pairs are sorted to prioritize **higher resolution first**:

1. descending $r$
2. for equal $r$, descending $k_x$

This ordering ensures that we start with the most informative configuration and only reduce resolution if necessary.

Or in other words we create a matrix of $(r,k)$ pairs where columns represent the $k$ value grid descending from left to right, and rows represent the $r$ value grid descending from top to bottom. We evaluate the pairs in that matrix row, after row, and within each row from left to right.

---

### 4. Bootstrap-based stability assessment (no permutations yet)

For each parameter pair in the sorted list:

1. Run the JSCompTest **without permutation testing**
2. Apply bootstrap resampling of residuals within neighborhoods
3. Compute the global JSD and its confidence interval (CI)

Bootstrapping is used **only** to assess numerical stability, not significance.

---

### 5. Plateau detection and stopping criterion

As parameter pairs are processed in descending order:

- Track the overlap of confidence intervals between consecutive parameter pairs
- Stop when:
  - CI overlap exceeds a threshold (e.g. $>90\%$ overlap), **and**
  - the relative change in mean JSD is below a tolerance (e.g. $<5\%$)

This indicates that further reduction in $(r, k_x)$ no longer changes the estimated effect size in a meaningful way.

The first parameter pair at which this plateau is reached is selected as the **final operating point**.

---

### 6. Permutation testing at the selected resolution

Only **after** a stable $(r, k_x)$ has been selected:

- Perform permutation-based null hypothesis testing
- Use the fixed histogram geometry and neighborhood definitions
- Compute empirical p-values

This avoids unnecessary permutation costs for unstable or suboptimal parameter choices.

---

### 7. Efficient permutation via Fisher–Yates shuffling with PCG RNG

Permutation tests rely on shuffling study labels **without replacement**.

We implement Fisher–Yates shuffling using a fast, reproducible PCG-based RNG.

#### Fisher–Yates shuffle (Fortran)

```fortran
do i = n, 2, -1
  j = pcg_rand_bounded(i)   ! uniform integer in [1, i]
  call swap(labels(i), labels(j))
end do
```
or in more detail:
```fortran
subroutine fisher_yates_shuffle_int(a, n, rng_state)
  implicit none
  integer, intent(inout) :: a(:)
  integer, intent(in)    :: n
  integer(kind=8), intent(inout) :: rng_state
  integer :: i, j, tmp

  do i = n, 2, -1
     j = 1 + int( rand_u01(rng_state) * real(i) )  ! j in [1,i]
     tmp = a(i); a(i) = a(j); a(j) = tmp
  end do
end subroutine
```

PCG RNG module (vendored)
Self-contained Fortran module (e.g. pcg32.f90)

One RNG state per thread or task

Deterministic across gfortran, ifx, and nvfortran

No reliance on random_number

This provides:

fast permutation testing

reproducibility

compiler-independent behavior

##### Summary

This procedure ensures that:

parameter choices are data-driven

effect sizes are numerically stable

permutation tests are computationally efficient

runtime scales to large cohorts (e.g. TCGA)

The JSCompTest thus becomes both statistically principled and HPC-ready.

##### Asis' wording

Store for each residual the following information in a two dimensional int Array:
- residual index enabling us to get to the original real residual value. This can be several (multi dimensional) int indices, if necessary
- study id (index)
- neighborhood id (index)
- histogram bin id (index)

We only reshuffle through permutation (_without_ replacement) the study labels. Then we can recompute the (study, neighborhood) specific histrograms quickly from the above data structure. 

##### ChatGPT wording

_(Take with a grain of salt, please.)_

For each mean-expression neighborhood $j$, we store residual information in a **token-based representation**.

Let $N_j$ denote the total number of residual samples pooled into neighborhood $j$ across all studies.

We define the following arrays:

- `bin_idx(1:N_j)`  
  Integer array storing the histogram bin index for each residual token.  
  Values are in $\{1,\dots,M\}$ and **remain fixed across all permutations**.

- `lab(1:N_j)`  
  Integer array storing the study label for each residual token.  
  Values are in $\{1,\dots,K\}$, where $K$ is the number of studies.  
  **This array is permuted during permutation tests.**

Using this representation, histogram counts are reconstructed as:

$$
\text{counts}(s, m) = \#\{\, t \mid \text{lab}(t) = s \;\land\; \text{bin\_idx}(t) = m \,\}
$$

for study $s \in \{1,\dots,K\}$ and bin $m \in \{1,\dots,M\}$.

##### Advantages

- No residual values need to be resampled or recomputed
- Permutation reduces to shuffling `lab(·)` without replacement
- Histogram geometry and bin assignments remain invariant
- Memory layout is compact and cache-friendly
- Ideal for high-performance Fortran implementations

This layout supports fast, reproducible permutation testing while preserving the exact empirical residual structure within each neighborhood.

*Migrated from GitLab: https://gitlab.rlp.net/a.hallab/tensor-omics/-/work_items/125*

qwerdenkerXD:
Implements [this Issue](https://gitlab.rlp.net/a.hallab/tensor-omics/-/work_items/125)

Additionally, there are the following changes:
- the test runner now uses fpm as well for builds. This simplifies the script and allows compilation of only changed tests, just as we already have with the build script. To enable this, the `--reuse-mod-files` flag needs to be set for the tests runner. By default, the tests will be all recompiled as usual. This is because it happens that if only src changes, fpm doesn't seem to relink the tests properly, so they use the old state. I don't know yet why exactly, so this is why the usual recompilation of all test files is the default.
- tox_errors now supports argument specific error codes. The codes have the pattern `<arg_pos><code>`, which makes them seemlessly compatible with our current implementation, as the optional `arg_pos` will be just zero there, which means an internal non-argument-related error. The error handling can be simply updated step-wisely to have this extra information, which is really helpful for debugging
- the asserts module now has an extra operator overload. It enables string concatenation with integers, which is something I really missed sometimes when writing assertion messages for assertions inside loops, where the failing iteration is good to know

Closes #125, #126


src/tox_data_integration_js_comp_test.F90
```diff
@@ -0,0 +1,1319 @@
...
+                            candidates_n_points_n_neighbors(1, i_candidate) = point_candidate
+                            i_candidate = i_candidate + 1
+                        end if
+                    end do
+                end if
+
+                n_points_candidate_real = n_points_candidate_real * GAMMA
+            end associate
+        end do
+        max_n_neighbors_candidate = candidates_n_points_n_neighbors(2, 1)
```
qwerdenkerXD: I think I will source this out into a helper to make it testable

src/tox_data_integration.F90
qwerdenkerXD, line 1: I will add the C wrappers here.

Regarding the submodule change. This way here is better, as it is less redundant. The problem with submodules was that I really needed to define the explicit interfaces and duplicate the Ford comments, as Ford took the comments from the interfaces.
I don't fully understand the actual need of submodules, as ***all*** used names in a module, like `int32`, `real64`, or the imported subroutines can  be used from other modules as well. So even without the interfaces here, one could just import the used modules from `tox_data_integration_js_comp_test`.
I only decided to add the interfaces for documentation reasons, so the Ford page has something to show, and the actual purpose of the module is clear.

README.md
qwerdenkerXD, line 1: GitHub doesn't seem to support opening threads unrelated to code, so I will leave it here:

I will merge main once Jitu's branch is merged

src/tox_data_integration_js_comp_test.F90
```diff
@@ -0,0 +1,1534 @@
...
+                call compute_divergence_per_reference_point_helper(tmp_pmfs(:, :, i_study), mean_pmf, n_points, n_bins, tmp_js_divergences(:, i_study))
+                call compute_weighted_global_divergence_helper(tmp_js_divergences(:, i_study), n_points, included_n_reps(:, i_study), mean_pmf_included_n_reps, tmp_global_js_divergence(i_study), tmp_weights(:, i_study))
+
+                if (tmp_global_js_divergence(i_study) >= global_jsd_observed(i_study)) then
+                    p_values(i_study) = p_values(i_study) + 1.0_real64
+                end if
+            end do
+        end do
+
+        if (n_permutations == 0) return
```
VivianBass, line 1302: shouldn't we use actual_n_permutations?

src/tox_data_integration_js_comp_test.F90
```diff
@@ -0,0 +1,1534 @@
...
+            ! For new pmfs from resampling, run the pipeline to determine the JSD
+            do concurrent (i_study = 1:n_studies) shared(tmp_pmfs, tmp_mean_pmf, n_points, n_bins, tmp_js_divergences, included_n_reps, mean_pmf_included_n_reps, tmp_global_js_divergence, tmp_weights, bootstrapping_top_k_jsds, n_bootstrapping_top_k_jsds)
+                call compute_divergence_per_reference_point_helper(tmp_pmfs(:, :, i_study), tmp_mean_pmf, n_points, n_bins, tmp_js_divergences(:, i_study))
+                call compute_weighted_global_divergence_helper(tmp_js_divergences(:, i_study), n_points, included_n_reps(:, i_study), mean_pmf_included_n_reps, tmp_global_js_divergence(i_study), tmp_weights(:, i_study))
+
+                ! Keep highest and lowest values
+                call bottom_k_heap_push(bootstrapping_top_k_jsds(:, 1, i_study), n_bootstrapping_top_k_jsds, tmp_global_js_divergence(i_study))
+                call top_k_heap_push(bootstrapping_top_k_jsds(:, 2, i_study), n_bootstrapping_top_k_jsds, tmp_global_js_divergence(i_study))
+            end do
+        end do
```
VivianBass, line 1380: In this section, the resampling is performed using `mean_pmf_counts` as the source for each study, and then the mixture is recalculated. However, according to the final agreements in Issue #126, the expected workflow was to resample each study independently from its own data and then rebuild the mixture from those resampled studies in every iteration. Am I missing something? Perhaps this was discussed in a meeting and I've lost the context, but the current implementation seems to resample from the global mean instead of the individual studies.

src/tox_data_integration_js_comp_test.F90
```diff
@@ -0,0 +1,1534 @@
...
+        exceeds_min_CI_overlap = 0_int32
+        do concurrent (i_study = 1:n_studies) shared(confidence_interval, best_candidate_pair_confidence_interval, succeeding_ci_overlap) reduce(+:exceeds_min_CI_overlap)
+            associate (&
+                ci_min => confidence_interval(1, i_study),&
+                ci_max => confidence_interval(2, i_study),&
+                best_ci_min => best_candidate_pair_confidence_interval(1, i_study),&
+                best_ci_max => best_candidate_pair_confidence_interval(2, i_study)&
+            )
+                ! Check overlap
+                ! IMPORTANT: best interval needs to be second argument, as the denominator will be the range of first interval
```
VivianBass, line 1201: Again this is only a comment because it's been a while and I don't remember. Here we use an asymmetric fractional overlap. According to the discussion in Issue 125, we agreed on using Intersection over Union to measure CI stability. 
Did we decide later to change it?

src/tox_data_integration_js_comp_test.F90
qwerdenkerXD: The resampling works like this:
1. collect all residuals across studies for one neighborhood (or directly the assigned bin here)
2. take this as pool to resample the individual neighborhoods
3. recreate mean pmf

So why am I using mean_pmf_counts as source? It is equivalent to `sum(pmf for pmf in pmfs)`, meaning exactly the described above

src/tox_data_integration_js_comp_test.F90
qwerdenkerXD: Was it really discussed or more proposed? My problem with union is that if the current best interval includes the new one, like `bestmin<newmin<newmax<bestmax`, the overlap will be worse compared to the union, but actually the new interval is better. This is the bad thing with symmetry, we don't know which side caused the good/bad value

src/tox_data_integration_js_comp_test.F90
```diff
@@ -0,0 +1,1534 @@
...
+        M_DEFAULT_VAL(min_count_per_mean_bin, actual_min_count_per_mean_bin, 5_int32)
+        M_DEFAULT_VAL(min_neighbor_overlap, actual_min_neighbor_overlap, 0.1_real64)
+        M_DEFAULT_VAL(succeeding_ci_overlap, actual_succeeding_ci_overlap, 0.9_real64)
+
+        n_gene_means = size(gene_means, kind=int32)
+        n_pool = find_last_non_nan(gene_means, gene_means_perm_all, n_gene_means)
+
+        best_candidate_pair_confidence_interval = -1.0_real64
+        best_params_CI_i_candidate_n_points_n_neighbors = 1
+        best_params_CI_i_candidate_n_points_n_neighbors = 1
```
VivianBass, line 774: Duplicated lines

src/tox_data_integration_js_comp_test.F90
VivianBass: To answer your question on why I mentioned it's being used as the source:

In the code, `mean_pmf_counts` is passed as the second argument to `random_multinomial`. This means that when you draw the resampled counts for Study A, Study B, etc., they are all being generated from the same probability distribution (the mean).

Now, as I understand it, here is how the steps should look:

1. Requested Workflow:

* Source: Take Study A's original counts in that neighborhood $\rightarrow$ `random_multinomial` $\rightarrow$ Result: Study A*.
* Source: Take Study B's original counts in that neighborhood $\rightarrow$ `random_multinomial` $\rightarrow$ Result: Study B*.
* Reconstruction: After resampling all studies independently, call `create_mean_pmf_only_helper` to build the new mixture ($M^*$) for this specific iteration.
* Comparison: Compare each $S_i^*$ against this new $M^*$.

2. Current Implementation:

* Source: Take the Mixture (`mean_pmf_counts`) $\rightarrow$ `random_multinomial` $\rightarrow$ Result: Study A* and Study B*.
* Reconstruction: Re-average them (which will likely just point back to the original mean).

The key difference is that in the first case, the Mixture is a result that fluctuates based on individual study variance. In the current code, the Mixture is the source, which 'smooths' the variance before we even start the iteration.

Since the Issue explicitly said: 'do resample each study... and do not resample the mixture', I think we should change the source of `random_multinomial` from `mean_pmf_counts` to the individual study counts. 

What do you think? And please if I'm saying something wrong let me know. I'm just basing my comments in the issues so I know that probably I'm missing something.

PR #135, opened 2026-05-13
122 parallelize codebase

### LINKED ISSUE #122: Parallelize codebase
Replace standard loops with `do concurrent` in all our codebase.

Apply only to root-level loops; avoid nested parallelization.

**Assignees:** @qwerdenkerXD

*Migrated from GitLab: https://gitlab.rlp.net/a.hallab/tensor-omics/-/work_items/121*

qwerdenkerXD:
- Replaced loops by concurrent loops where applicable
- Refactored/Polished whole codebase
- Improved test runner to use fpm, added general information about command line options of `test_runner.sh` and `build.sh`
- improved handling of external libs (loess_netlib)
- colorized test runner outputs and build script messages
- improved Fortran assertions formatting
- returning to one static `fpm.toml`. With fpm version 0.13 compiler specific features are possible, so specific link configs, which makes the dynamic generation unnecessary and we can follow the more conventional way.
- merged `ford.yml` into `fpm.toml`. The Ford documentation actually recommends the use of the `fpm.toml` for that, because of the more forgiving `toml` syntax. As we have the `fpm.toml` back, why not using it for that
- added module hierarchy to src like `src/tox/data/tox_data_integration.F90`
- all optimization flags are now part of the `--max-performance` option for the build/test_runner scripts. Without this option, the only flag is `-fPIC`
    - the GitHub Workflow runs both variants, with and without this option to ensure that it runs with optimizations, but also without (so no errors shadowed by optimizations)
- merged general_documentation into this branch, so once this PR is being merged, the main branch has the manual as well
- added `-I.` to compilation flags -> allows better macro inclusion: `#include <src/macros.h>` or `#include <src/f42/serde/macros.h>`, so we can have multiple macro files without ambitious relative include paths
- added `authors.h`. It is included in `src/macros.h` and allows documenting the author in Ford comments, like
  ```fortran
  !> AUTHOR_AARON_SCHROEDER
  !| This cool routine bla
  subroutine bla
  ```
  This will expand to the Ford author meta tag and liks to the GitHub profile of Aaron here

Closes #122


python/test/mod_test_tca_normalization.py
```diff
@@ -15,134 +15,92 @@
...
-        result = tox_normalize_variable_timeseries(v)["v_norm"]
-        assert np.all(v == 0.0), "Zero vector should remain zero"
-        print("Zero vector case passed")
-    except RuntimeError:
-        print("Zero vector raised error")
-
-    print("\n" + "="*50)
-    print("✓ ALL tox_normalize_variable_timeseries TESTS PASSED")
-    print("="*50)
+    v, status = tox_normalize_variable_timeseries(v).values()
```
schroederaaron, line 70: If i remember correctly more than one return value should always be returned as a dictionary. If so, please change.

python/error_handling.py
```diff
@@ -1,6 +1,8 @@
 #> tox_helper: throw error in error case
-def check_err_code(ierr: int) -> None:
-    if ierr == 0:
+def check_err_code(ierr: int, arg_pos_map: dict = {}) -> None:
+    arg_pos, code = divmod(ierr, 10000)
```
schroederaaron, line 3: What is the purpose of this? `arg_pos` will always be zero and `code` will always be equal to `ierr`, wont it?

misc/tox_manual/error_handling.tex
```diff
@@ -0,0 +1,60 @@
...
+
+    \multicolumn{2}{l}{\textbf{3xx: Memory Errors}} \\[3pt] 
+    301 & Memory allocation failed. \\ 
+    302 & Null pointer reference encountered. \\[6pt]
+
+    \multicolumn{2}{l}{\textbf{5xxx: Fortran Runtime / Unit State Errors}} \\[3pt] 
+    5002 & Fortran runtime error: unit not open or not connected. \\[6pt]
+
+    \multicolumn{2}{l}{\textbf{9xxx: Internal / Unknown Errors}} \\[3pt] 9001 & Internal
+    error: unexpected state or invariant violated. \\ 9999 & Unknown error (fallback
```
schroederaaron: general question regarding the 9999, is there a case where that code comes back? I don't think it is something fortran returns, is it? So if we do not manually set it (what we don't do) this will never occur, will it?

misc/tox_manual.tex
```diff
@@ -0,0 +1,54 @@
...
+
+\input{tox_manual/methods_and_theory}
+\input{tox_manual/error_handling}
+
+\section{Implementation Notes}
+
+This section details the practical pipeline for using the TOX library to read and
+process tabular data. The workflow is consistent across Fortran, Python, and R,
+starting with reading a CSV file into a staging area of strings and then converting
+columns to their proper data types.
```
schroederaaron: Do we have the implementation of the `read_columns_as_bands` (or whatever the actual name was)? I remember that some refactoring etc was planned but don't fully remember if the reading a csv function was abolished or if we have an implementation of that

rcpp/test/mod_test_which.R
schroederaaron, line 1: Why exactly are we exposing the fortran implementation of `which` if R has a native `which` function? For python we could use the `np.where` function, so we wouldn't need to expose the function at all, according to the coding guides.

rcpp/error_handling.R
```diff
@@ -1,45 +1,74 @@
 #> tox_helper: throw error in error case
-check_err_code <- function(ierr) {
-  if (ierr == 0) return(invisible(NULL))
-  msg <- switch(as.character(ierr),
+check_err_code <- function(ierr, arg_pos_map = setNames(list(), character())) {
+  # R equivalent of Python's divmod
+  arg_pos <- ierr %/% 10000
+  code    <- ierr %% 10000
```
schroederaaron, line 5: Same as in python, why the `%%10000`?

src/f42/serde/arrays/deserialize/f42_serde_arrays_deserialize_char.F90
```diff
@@ -0,0 +1,166 @@
+#include <src/macros.h>
+
+!> Module for deserializing character arrays from files
+module f42_serde_arrays_deserialize_char
```
schroederaaron, line 4: why did they get renamed to `serde`?

schroederaaron (changes requested):
Did not check all files yet. Some tex files and fortran code is still open.

misc/tox_manual/error_handling.tex
qwerdenkerXD: well, you were the one adding this code to tox_errors, so tell me

python/error_handling.py
qwerdenkerXD: nope, validation routines now accept argument positions that will be included in the error code, like `<arg_pos><4digits_ierr>`

python/test/mod_test_tca_normalization.py
qwerdenkerXD: it is? I unpack the .values() output of the returned dict

rcpp/test/mod_test_which.R
qwerdenkerXD: I know, this function is quite useless and not even used anywhere in our code. It's some relic from the very beginning. @VivianBass should I remove it?

rcpp/error_handling.R
qwerdenkerXD: see tox_errors

src/f42/serde/arrays/deserialize/f42_serde_arrays_deserialize_char.F90
qwerdenkerXD: for structuring the src directory. `serde` means serialization and deserialization. As we will add json, and maybe csv as well, the serde directory will be a good place for them.

misc/tox_manual.tex
qwerdenkerXD: well, I can also remove the manual from this branch and we create a separate merge request for merging the general_documentation into main.

I just merged it into this one, not touching any of the files

python/error_handling.py
schroederaaron: Ahhh i see, so we no longer set ierr to e.g. 301 but to say 40301 to say argument number 4 caused error 301?

misc/tox_manual/error_handling.tex
schroederaaron: Well yeah then that one is on me. I don't think there is a case where we need it, nor is it returned by fortran. I mean i could not really think of a case where we return an error but can not tell where it occured, so i think we should remove it

misc/tox_manual.tex
schroederaaron: Might be better because there are currently like 260 files changed with a lot of content not really related to the original MR i think. What do you think @VivianBass ?

python/test/mod_test_tca_normalization.py
schroederaaron: Ah yes of cause, my bad you use the .values(), didn't realize it. I thought you return two single values from the function.

src/f42/f42_utils.F90
```diff
@@ -0,0 +1,1699 @@
...
+            !! Error code
+
+        integer(int32) :: i_dim
+        real(real64) :: theta, dot_product, norm1, norm2, norm_product
+
+        call set_ok(ierr)
+
+        dot_product = 0.0_real64
+        norm1 = 0.0_real64
+        norm2 = 0.0_real64
```
schroederaaron: `norm2` is also the name of a function, not sure if that could cause trouble on some compilers

src/f42/f42_utils.F90
```diff
@@ -0,0 +1,1699 @@
...
+        else if (ieee_is_nan(b)) then
+            real_greater = .false.
+        else
+            real_greater = (a > b)
+        end if
+    end function real_greater
+
+    !> AUTHOR_VIVIAN_BASS
+    !| Finds the indices of the true values in a logical mask.
+    pure subroutine which(mask, n, idx_out, m_max, m_out, ierr)
```
schroederaaron, line 1094: the argument naming does not align with our latest guidelines, does it?

src/f42/f42_utils.F90
```diff
@@ -0,0 +1,1699 @@
...
+    !|
+    !| Implements:
+    !|   P(d) = ( #{di in D | di >= d} + c ) / ( |D| + c )
+    !|
+    !| Because distances are non-negative, a one-sided upper-tail empirical p-value is used.
+    !|
+    !| Assumptions / preconditions:
+    !| - sorted_rdi(1:n_genes) contains the empirical distribution D.
+    !| - If invalid RDIs exist (negative), they should already be mapped to 0 in the distribution
+    pure subroutine compute_empirical_p_values(n_genes, rdi, sorted_rdi, perm, p_values, c_const)
```
schroederaaron: Do we have tests for this function that display the distribution of resulting pvalues? I think we have a slight statistical issue here, as the values we are testing are tested against their own distribution, so this is more of a ranking rather than an actual pvalue. The results will in every case be uniformly distributed and (statistically spoken) can not be named pvalues as their is no real null we test against. The null we test against will contain the exact same signal we are possibly trying to find. I strongly suggest a renaming of the function, as the return is a ranking of observed distances compared to other observed distances.

src/f42/f42_utils.F90
```diff
@@ -0,0 +1,1699 @@
...
+            end if
+
+            p_values(i) = (real(count_ge, real64) + c_const)/denom
+        end do
+
+    end subroutine compute_empirical_p_values
+
+    !> AUTHOR_VIVIAN_BASS
+    !| First position pos in [1..n] such that sorted_rdi(perm(pos)) >= x. Returns n+1 if none.
+    pure integer(int32) function lower_bound_ge(vals, p, n, x) result(pos)
```
schroederaaron: please rename the arguments

python/error_handling.py
qwerdenkerXD: in the optimal case, yes. It still works without a set argument position, as an inset position means position zero, which is treated as internal error.

The position is just good sugar for debugging

src/tox/tox_euclidean_distance.F90
```diff
@@ -0,0 +1,198 @@
...
+        real(real64), dimension(n_elements), intent(in) :: vec2
+            !! Second expression vector
+        real(real64), intent(out) :: result
+            !! Output scalar distance
+
+        integer(int32) :: i_element
+        real(real64) :: sum_squared_diff
+
+        sum_squared_diff = 0.0_real64
+        ! GFORTRAN BUG: do concurrent (i_element = 1:n_elements) shared(vec1, vec2) reduce(+:sum_squared_diff)
```
schroederaaron, line 57: Just out of curiosity, what is the bug here?

src/tox/tox_paralog_analysis.F90
```diff
@@ -239,23 +267,24 @@ pure subroutine detect_patterns(ancestor, genes, n_genes, n_dims, pattern, filte
         do subset_size = 2, max_subset_size
             n_new_active_masks = 0_int32
             do while (n_active_masks > 0)
-                call take_active_mask_helper(work_arr_paralog_subsets, n_mask_chunks, n_paralog_subsets, n_results, n_active_masks, n_new_active_masks, active_mask, ierr)
+                call take_active_mask_helper(work_arr_paralog_subsets, n_mask_chunks, n_paralog_subsets, n_results, n_active_masks, n_new_active_masks, tmp_active_mask, ierr)
```
schroederaaron, line 275: missing error code mapping, e.g. the `generate_subset_helper` will return `INVALID_INPUT` for argument 4, caused by `pattern` which will then be returned as `INVALID_INPUT` for argument 4 from the main routine, indicating an error in `n_dims`.

schroederaaron: goes for this call as well as for the `generate_subsets_helper` call, i just can not comment on that line for some reason.

src/tox/tox_paralog_analysis.F90
```diff
@@ -476,15 +507,15 @@ pure subroutine take_active_mask_helper(subsets, n_mask_chunks, n_subsets, n_res
...
-        call validate_dimension_size(n_subsets, ierr)
-        call validate_in_range_int(n_active_masks, ierr, min=1_int32)
-        call validate_in_range_int(n_results, ierr, min=0_int32)
-        call validate_in_range_int(n_new_active_masks, ierr, min=0_int32)
+        call validate_dimension_size(n_mask_chunks, ierr, arg_pos=2_int32)
+        call validate_dimension_size(n_subsets, ierr, arg_pos=3_int32)
+        call validate_in_range_int(n_active_masks, ierr, min=1_int32, arg_pos=5_int32)
+        call validate_in_range_int(n_results, ierr, min=0_int32, arg_pos=4_int32)
+        call validate_in_range_int(n_new_active_masks, ierr, min=0_int32, arg_pos=6_int32)
         if (is_err(ierr)) return
```
schroederaaron, line 525: Isn't this already checked by the caller, as we are in the `_helper`, i mean it is only like single value validations, but i think it will cause error code mapping issues in the caller function.

src/tox/tox_paralog_analysis.F90
```diff
@@ -521,15 +553,15 @@ pure subroutine add_to_results_helper(subsets, n_mask_chunks, n_subsets, n_resul
...
+            !! Error code
 
         call set_ok(ierr)
 
-        call validate_dimension_size(n_mask_chunks, ierr)
-        call validate_dimension_size(n_subsets, ierr)
-        call validate_in_range_int(n_active_masks, ierr, min=0_int32)
-        call validate_in_range_int(n_results, ierr, min=0_int32)
-        call validate_in_range_int(n_new_active_masks, ierr, min=0_int32)
+        call validate_dimension_size(n_mask_chunks, ierr, arg_pos=2_int32)
```
schroederaaron, line 566: same as above

src/tox/tox_relative_axis_plane_tools.F90
```diff
@@ -0,0 +1,705 @@
...
+        integer(int32), intent(out) :: ierr
+            !! Error code
+
+        call set_ok(ierr)
+
+        call validate_dimension_size(n_dims, ierr, arg_pos=2_int32)
+        call validate_dimension_size(n_fields, ierr, arg_pos=3_int32)
+        call validate_dimension_size(n_selected_fields, ierr, arg_pos=5_int32)
+        call validate_all_in_range_real(fields, size(fields, kind=int32), ierr, arg_pos=1_int32)
+        if (count(fields_selection_mask, kind=int32) > n_selected_fields) call set_err(ierr, ERR_INVALID_INPUT, arg_pos=4_int32)
```
schroederaaron: Earlier in `omics_field_RAP_projection` you set the error code for the size argument, not for the array argument in this case. Both is correct i think, but you should decide on one for consistency.

src/macros.h
```diff
@@ -14,3 +14,8 @@
 #define M_NAN ieee_value(1.0_real64, ieee_quiet_nan)
 #define M_NEG_INF ieee_value(1.0_real64, ieee_negative_inf)
 #define M_POS_INF ieee_value(1.0_real64, ieee_positive_inf)
+
+#define M_GENE_TO_FAM_SENTINEL 0_int32
```
schroederaaron, line 26: Is this actually necessary? I mean sure if we later decide at some point that we will introduce another value for this, then yes. But i don't really think that will be the case

test/mod_test_binary_search.F90
```diff
@@ -0,0 +1,114 @@
...
+module mod_test_binary_search
+    use asserts
+    use, intrinsic :: iso_fortran_env, only: real64, int32
+    use, intrinsic :: ieee_arithmetic, only: ieee_value, ieee_quiet_nan
+    use f42_utils
+    use tox_errors
+    use test_suite, only: test_case
+    implicit none
+
+    real(real64), parameter :: TOL = epsilon(1.0_real64)
```
schroederaaron: i know i wrote this but i think it is an artifact from a quite long time ago, should be replaced with our regular 1e-12 tolerance probably

test/mod_test_clock_hand_angles.F90
```diff
@@ -0,0 +1,422 @@
...
+    use tox_relative_axis_plane_tools
+    use, intrinsic :: iso_fortran_env, only: real64
+    use test_suite, only: test_case
+    use tox_errors
+    implicit none
+    public
+
+    ! Mathematical constants
+    real(real64), parameter :: PI = 3.141592653589793_real64
+    real(real64), parameter :: TOL = 1e-12_real64
```
schroederaaron: both can be imported from f42_utils i think

test/mod_test_trajectory_contribution_analysis.F90
```diff
@@ -1,28 +1,28 @@
...
-    use, intrinsic :: ieee_arithmetic, only: ieee_value, ieee_quiet_nan
+    use, intrinsic :: ieee_arithmetic, only: ieee_value, ieee_quiet_nan, ieee_positive_inf
     use tox_trajectory_contribution_analysis
     use tox_errors
     use tox_trajectory_normalization
     use test_suite, only: test_case
     implicit none
 
-    
     real(real64), parameter :: TOL = epsilon(1.0_real64)
```
schroederaaron: probably also import from f42_utils

schroederaaron (changes requested):

src/tox/tox_euclidean_distance.F90
qwerdenkerXD: in gfortran 16.1.1 on my local archlinux WSL, I get this when using the mentioned concurrent variant:
```
Using compiler: gfortran                                                                                                                                                                                                           
config.F90                             done.                                                                                                                                                                                       
tox_errors.F90                         done.                                                                                                                                                                                       
tox_conversions.F90                    done.                                                                                                                                                                                       
safeguard.F90                          done.                                                                                                                                                                                       
f42_utils.F90                          done.                                                                                                                                                                                       
f42_serde_utils.F90                    done.                                                                                                                                                                                       
tox_clustering.F90                     done.                                                                                                                                                                                       
tox_euclidean_distance.F90             failed.                                                                                                                                                                                     
[ 17%] Compiling...                                                                                                                                                                                                                
f951: internal compiler error: in gfc_resolve_forall, at fortran/resolve.cc:12805                                                                                                                                                  
0x268c7d8 diagnostics::context::diagnostic_impl(rich_location*, diagnostics::metadata const*, diagnostics::option_id, char const*, __va_list_tag (*) [1], diagnostics::kind)                                                       
        ???:0                                                                                                                                                                                                                      
0x268ede7 internal_error(char const*, ...)                                                                                                                                                                                         
        ???:0                                                                                                                                                                                                                      
0x7220ba fancy_abort(char const*, int, char const*)                                                                                                                                                                                
        ???:0
0x841e47 gfc_resolve_code(gfc_code*, gfc_namespace*)
        ???:0
0x841cec gfc_resolve_code(gfc_code*, gfc_namespace*)
        ???:0
0x8450bb gfc_resolve_blocks(gfc_code*, gfc_namespace*)
        ???:0
0x841e47 gfc_resolve_code(gfc_code*, gfc_namespace*)
        ???:0
0x821e5d gfc_parse_file()
        ???:0
/usr/lib/gcc/x86_64-pc-linux-gnu/16.1.1/f951 ././src/tox/tox_euclidean_distance.F90 -cpp=/tmp/ccMXtmNB.fii -quiet -I . -I build/gfortran_D10FC5326AE348E7 ././src/tox/tox_euclidean_distance.F90 -quiet -dumpdir build/gfortran_D10FC5326AE348E7/tensor-omics/ -dumpbase src_tox_tox_euclidean_distance.F90.F90 -dumpbase-ext .F90 -mtune=generic -march=x86-64 -Werror=implicit-interface -fPIC -fimplicit-none -ffree-form -J build/gfortran_D10FC5326AE348E7 -fintrinsic-modules-path /usr/lib/gcc/x86_64-pc-linux-gnu/16.1.1/finclude -fpre-include=/usr/include/finclude/math-vector-fortran.h -o /tmp/ccUBzQ9i.s
Please submit a full bug report, with preprocessed source (by using -freport-bug).
Please include the complete backtrace with any bug report.
See <https://gitlab.archlinux.org/archlinux/packaging/packages/gcc/-/issues> for instructions.
<ERROR> Compilation failed for object " src_tox_tox_euclidean_distance.F90.o "
<ERROR> stopping due to failed compilation
STOP 1
Error: Build with fpm failed - Exit code: 1
```

So I don't know the actual error here. gfortran has some processing issues here

src/f42/f42_utils.F90
qwerdenkerXD: too bad you didn't review the loess_integration MR. Anyway, this is actually an implementation of binary search. I did not check that too much, but I think it can be discarded completely, so the bin search routines are being used at the respective place

src/f42/f42_utils.F90
qwerdenkerXD: @VivianBass

src/macros.h
qwerdenkerXD: it is not too unlikely. Maybe we decide for `-1` at some point to have a consistent sentinel value, or just to ensure it is not a valid index in any language.

Anyway, I think it's not too bad practice.

@VivianBass your opinion? Undoing this is just a single `sed` call, so tell me what to do

src/tox/tox_relative_axis_plane_tools.F90
qwerdenkerXD: true, I don't remember why I decided differently here

test/mod_test_binary_search.F90
qwerdenkerXD: no, this one is from me. You wrote the mod_test_bst

src/f42/f42_utils.F90
schroederaaron: I remember working with the loess integration branch... So, not to sure if it can be simply removed due to possible downstream dependencies, but i think it is worth checking if it can be easily removed. It looks like a binary search to me, so i think a more clear `binary_search(values, values_perm, target, position, ierr)` or similar makes more sense so it is easier to use. What do you think @VivianBass ?

src/tox/tox_euclidean_distance.F90
schroederaaron: Huh, interesting

src/macros.h
schroederaaron: You are right. I think i'd leave it as it is

src/tox/tox_trajectory_normalization.F90
```diff
@@ -1,167 +1,170 @@
...
+    !| Normalize all factors in a single trajectory independently across time
+    !| Input: `trajectory(n_factors, n_timepoints)` for ONE sample/entity
     pure subroutine normalize_single_trajectory(trajectory, trajectory_norm, n_factors, n_timepoints, ierr, status)
         integer(int32), intent(in) :: n_factors
-        !! Number of factors/variables
+            !! Number of factors/variables
         integer(int32), intent(in) :: n_timepoints
-        !! Number of time points
+            !! Number of time points
         real(real64), intent(in) :: trajectory(n_timepoints, n_factors)
```
VivianBass, line 72: In the documentation we say n_factors,n_timepoints but here is the other way around. Which one is the correct for efficiency?

src/tox/tox_trajectory_normalization.F90
```diff
@@ -1,167 +1,170 @@
...
+        M_ALLOCATE(tmp_series_norm(n_timepoints))
+
         ! Normalize each sample/entity independently
         do i_sample = 1, n_samples
             do i_factor = 1, n_factors
-                do i_timepoint = 1, n_timepoints
-                    temp_series(i_timepoint) = trajectories(i_factor, i_sample, i_timepoint)
+                do concurrent (i_timepoint = 1:n_timepoints) shared(tmp_series, trajectories, i_sample, i_factor)
+                    tmp_series(i_timepoint) = trajectories(i_factor, i_sample, i_timepoint)
                 end do
```
VivianBass, line 143: I think we are applying parallelization at a grain size that is too fine by opening the do concurrent loop on the time dimension. If time-series data, is quite small (e.g., 6 or 12), the CPU overhead required to spawn and synchronize threads will likely outweigh the mathematical benefits of the calculation itself. Furthermore, marking tmp_series as shared creates a latent race condition across iterations.

What if we invert the order of parallelization instead? We could move the do concurrent statement to the outermost loop (either for samples or factors) and keep the traversal across time points sequential within each individual thread. If we also declare tmp_series as local to each outer iteration, every thread will get its own private memory allocation on the stack, completely eliminating the race condition while properly leveraging multi-threading.

src/tox/tox_trajectory_contribution_analysis.F90
```diff
@@ -122,114 +195,151 @@ pure subroutine compute_p_values(local_contributions_observed, total_contributio
...
             end if
 
-            do i_timepoint = 1, n_timepoints
-                if (local_contributions_perm(i_timepoint, i_perm) >= local_contributions_observed(i_timepoint)) then
+            do concurrent (i_timepoint = 1:n_timepoints) shared(local_contributions_perm_test, i_perm, local_contributions_observed, local_p_values)
+                if (local_contributions_perm_test(i_timepoint, i_perm) >= local_contributions_observed(i_timepoint)) then
                     local_p_values(i_timepoint) = local_p_values(i_timepoint) + 1.0_real64
                 end if
             end do
         end do
```
VivianBass, line 269: I think we have the same overhead here, wouln't it better if we parallelize the permutation instead of the time? that way each permutation is independent from each other?

src/tox/tox_trajectory_normalization.F90
qwerdenkerXD: 1. Still, concurrent loops don't necessarily mean spawning threads
2. There is no race condition, each iteration `i_timepoint` assigns to `tmp_series(i_timepoint)`

src/tox/tox_trajectory_contribution_analysis.F90
```diff
@@ -79,41 +104,89 @@ subroutine perform_permutation_test(trajectories, n_factors, n_samples, n_timepo
...
 
         do i_perm = 1, n_permutations
-            call select_random_sample_helper(n_samples, sample_idx, random_sample, ierr)
-            if (is_err(ierr)) return
+            call select_random_sample_helper(n_samples, sample_idx, random_sample)
 
-            do i_timepoint = 1, n_timepoints
-                temp_dependent(i_timepoint) = trajectories(dependent_idx, random_sample, i_timepoint)
+            do concurrent (i_timepoint = 1:n_timepoints) shared(tmp_dependent, trajectories, dependent_idx, random_sample)
+                tmp_dependent(i_timepoint) = trajectories(dependent_idx, random_sample, i_timepoint)
```
VivianBass, line 188: Same here, wouldn't it better if we parallelize the permutations and explicitly pass random_sample and tmp_dependent as local variables to ensure true thread isolation?

src/tox/tox_trajectory_normalization.F90
qwerdenkerXD: I need to check, I think I only changed indentation in this module.

src/tox/tox_trajectory_contribution_analysis.F90
qwerdenkerXD: we can't, the permutation test uses the intrinsic randomizer, which is impure

src/tox/tox_trajectory_normalization.F90
VivianBass: OK, I understand, thanks for clarifying. But still, does it make sense for us to parallelize across the time dimension instead of factors or samples? If so, then please ignore all my other comments on this.

src/tox/tox_trajectory_contribution_analysis.F90
VivianBass: OK, given that refactoring or reimplementing that isn't in our immediate plans, it makes sense to leave it as is for now. However, if we had the opportunity to refactor it in the future, would it make more sense to parallelize across permutations instead of the time dimension? If so, and if it's something that would significantly improve performance, it might be worth keeping it in mind for a future issue.

src/tox/tox_normalization.F90
VivianBass, line 490: wouldn't this create a non deterministic behavior?

src/tox/tox_get_outliers.F90
VivianBass, line 199: same, wouldn't this create a non deterministic behavior?

VivianBass, line 285: same, wouldn't this create a non deterministic behavior?

src/tox/tox_euclidean_distance.F90
```diff
@@ -0,0 +1,198 @@
...
+
+    integer(c_int), intent(in), target :: n_genes
+        !! Total number of genes
+    integer(c_int), intent(in), target :: n_families
+        !! Total number of gene families
+    integer(c_int), intent(in), target :: n_elements
+        !! Expression vector dimension
+    real(c_double), dimension(n_elements, n_genes), intent(in), target :: genes
+        !! Gene expression matrix (n_elements × n_genes), column-major
+    real(c_double), dimension(n_elements, n_genes), intent(in), target :: centroids
```
VivianBass: shouldn't this be n_elements, n_families?
In the subroutine we have 
`  !! Family centroid matrix (n_tissues × n_families), column-major, `DISTANCE_SENTINEL` for unassigned genes`

src/tox/tox_errors.F90
```diff
@@ -0,0 +1,416 @@
...
+            call validate_all_in_range_real(distances, n*n, ierr, arg_pos, actual_min, max)
+            if (is_err(ierr)) return
+
+            do concurrent(i_col=1:n) shared(ierr, distances, arg_pos)
+                if (distances(i_col, i_col) /= 0.0_real64) then
+                    call set_err_once(ierr, ERR_INVALID_INPUT, arg_pos)
+                else
+                    do concurrent(i_row=1:n) shared(ierr, i_col, distances, arg_pos)
+                        if (distances(i_row, i_col) /= distances(i_col, i_row)) then
+                            call set_err_once(ierr, ERR_INVALID_INPUT, arg_pos)
```
VivianBass, line 416: race condition because they are non independent?

src/tox/tox_errors.F90
```diff
@@ -0,0 +1,416 @@
...
+        integer(int32), intent(in), optional :: arg_pos
+            !! Position of the validated argument that triggered the error, default: 0 -> not argument related
+        integer(int32), intent(in), optional :: sentinel
+            !! Optional sentinel value to allow additionally
+
+        integer(int32) :: i_element
+
+        if (present(array)) then
+            do concurrent(i_element=1:n_elements) shared(ierr, arg_pos, min, max, array, sentinel)
+                call validate_in_range_int(array(i_element), ierr, arg_pos, min, max, sentinel)
```
VivianBass, line 294: non deterministic behavior in ierr?

src/tox/tox_errors.F90
```diff
@@ -0,0 +1,416 @@
...
+        integer(int32), intent(in), optional :: arg_pos
+            !! Position of the validated argument that triggered the error, default: 0 -> not argument related
+        real(real64), intent(in), optional :: sentinel
+            !! Optional sentinel value to allow additionally
+
+        integer(int32) :: i_element
+
+        if (present(array)) then
+            do concurrent(i_element=1:n_elements) shared(ierr, arg_pos, min, max, array, sentinel)
+                call validate_in_range_real(array(i_element), ierr, arg_pos, min, max, sentinel)
```
VivianBass, line 375: non deterministic behavior in ierr?

qwerdenkerXD: true, I need to check that in the codebase. I think, at some place we have status codes that need to be switched to status code arrays as well

src/macros.h
VivianBass: I'd also leave this as it is; I think we might want a different value once we run more experiments.

src/tox/tox_euclidean_distance.F90
qwerdenkerXD: true, but not my fault. The original code was
```f90
  real(c_double), intent(in), target :: centroids(d, n_genes)
  !| Gene-to-family mapping (1-based indexing)
```

Do you know if anyone used this routine in analyses?

misc/tox_manual.tex
VivianBass: Good idea to remove the manual from this branch. We don't have the read_columns_as_bands, I think Alexander didn't finish it because he started his thesis and we closed the MR in gitlab.

src/tox/tox_euclidean_distance.F90
qwerdenkerXD: well, even if anyone used this routine, there shouldn't be an error, as the `distance_to_centroid` routine will still treat it as `(d, n_families)` because of explicit size

qwerdenkerXD: but I will change it

rcpp/test/mod_test_which.R
VivianBass: Well, I think this was for the read columns as bands? I don't remember, I think we can keep it in fortran and not expose it to R or C.

src/tox/tox_euclidean_distance.F90
VivianBass: I didn't mean it's your fault, I just pointed out the inconsistency.
No, I think everyone has used it in R (the previous wrapper).

Thanks for changing it! :)

src/tox/tox_trajectory_contribution_analysis.F90
qwerdenkerXD: ah, my bad, there is no randomness here, but the outer loop's iterations are not independent.
The actual calculation in this loop is `local_p_values(i_timepoint) = local_p_values(i_timepoint) + 1.0_real64`, so only dependent of `n_timepoints`.
If we parallelize the `n_permutations` loop, we introduce a race condition, as concurrent iterations might access the same timepoint in the output

src/f42/f42_utils.F90
VivianBass: This is a binary search lower bound type, and it is used in compute_empirical_p_values so we can't remove it just like that. Maybe we can use binary_search_insertion instead?

src/tox/tox_trajectory_contribution_analysis.F90
qwerdenkerXD: maybe something to keep in mind for a different time to refactor?

Doing this needs:
1. precomputing the randomly selected samples (can be stored in `total_contributions` to avoid an extra temporary)
2. extra workarrays `tmp_factor` and `tmp_dependent` for each permutation
3. parallelization friendly ierr handling: Currently, the `compute_contributions_helper` returns an `ierr` because of the baseline calculation
    - maybe introducing a non-validating helper for baseline calculation, while calling routines will validate baseline mode etc

VivianBass:
Why do we have a directory called "serde"? what does it mean?

src/f42/f42_utils.F90
qwerdenkerXD: They sound to be equivalent. I will replace it by the insertion function, in the lower_bound_ge tests as well

qwerdenkerXD:
> Why do we have a directory called "serde"? what does it mean?

Serialize/Deserialize. It is just for structuring the modules. All de-/serialization related modules will be part of it, as for arrays, JSON (when Flyer Json is merged), CSV (if we ever add it), maybe even HTML for the plot generation Asis wanted (generates the HTML for the D3 plots, by using the JSON serialization)

src/f42/f42_utils.F90
VivianBass: Thanks @schroederaaron, I think you have a good point. This is currently implemented exactly as Asis described in [this issue](https://gitlab.rlp.net/aaron.schroeder/bachelor-thesis-aaron/-/work_items/2#note_868262), where he refers to it as an *empirical p-value* and defines it as

$$
 P(d)=\frac{|{d_i\in D\mid d_i\ge d}|+c}{|D|+c}.
$$

I agree that these are not p-values in the strict inferential sense, since they are computed against the observed distribution itself rather than an independently generated null distribution.

My understanding of the issue was that the intention was to have a continuous measure of how extreme each distance is within the observed distribution, which is why Asis referred to them as *empirical* p-values rather than classical p-values.

I think it's worth clarifying the intended statistical interpretation before we decide whether the terminology or even the methodology should change. I can open a comment on the original issue and ask Asis to clarify what he intended. What do you think @schroederaaron @qwerdenkerXD ?

src/tox/data/integration/tox_data_integration_jsd.F90
```diff
@@ -245,12 +245,12 @@ pure module subroutine build_residual_histograms_helper(neighborhood_residuals,
         end do
```
VivianBass, line 252: I think the inner do concurrent over i_neighbor is unsafe because multiple iterations can update the same histogram cell for a fixed i_point.

For one outer iteration, i_point is constant, so all inner iterations write into counts(i_point, :). Since bin_idx is computed from residual values, different neighbors may map to the same bin_idx. That makes this statement a concurrent read-modify-write to the same location:

`counts(i_point, bin_idx) = counts(i_point, bin_idx) + 1
`
Isn't it a data race? Example: two neighbors both target counts(i_point, 6), both read 7, both write 8, and one increment is lost (expected 9). Or am I misunderstanding something?

src/tox/data/integration/tox_data_integration_jsd.F90
```diff
@@ -234,7 +234,7 @@ pure module subroutine build_residual_histograms_helper(neighborhood_residuals,
                         clamped_residual = clamp(neighborhood_residuals(i_rep, i_neighbor, i_point), min_val=-shared_residual_range, max_val=shared_residual_range)
 
                         ! assign bin to residual
-                        bin_idx = min(n_bins, int( (clamped_residual + shared_residual_range) / bin_width ) + 1)
+                        bin_idx = min(n_bins, int((clamped_residual + shared_residual_range)/bin_width) + 1)
```
VivianBass, line 244: Maybe we have a bug here? 
`bin_width` is zero when `shared_residual_range == 0`.
So, if all residuals in the neighborhood are identical, `shared_residual_range` is 0, so `bin_width = 0`. Then this produces a division by zero.

rcpp/test/mod_test_which.R
schroederaaron: I agree

src/f42/f42_utils.F90
schroederaaron: Yes i think that is a good idea @VivianBass , please do so.

src/tox/tox_errors.F90
schroederaaron: Yes there is a race condition on `ierr`. Even tho the result will be the same every time as there is only one error code that could be written, this still violates the formal requirement of independency. I think a simple `ANY` statement might be the better fit here as i am pretty sure it is highly optimized by the compiler.

src/tox/tox_errors.F90
schroederaaron: Same as above, i think a normal `ANY` might be a better fit

src/tox/data/integration/tox_data_integration_jsd.F90
schroederaaron: I agree, there is a race condition with potentially lost incrementations. Maybe you can declare the counts differently and then as `reduce(+:counts)`? Not fully sure if that works for arrays tho. Otherwise, make the counts atomic and use `ATOMIC_ADD` to get thread-safety. But i think atomic add with so many operations is very slow, so maybe just a sequential loop?

src/tox/tox_errors.F90
schroederaaron: An other option in general for this i realized is maybe introduce a flag? Like:

```
has_error = .false.
do concurrent ... reduce(.or.:has_error)
    ....
end do
if(has_error) call set_err(ERR_INVALID_INPUT)
```
I know it doesn't carries the actual error from every iteration but i think `ERR_INVALID_INPUT` is the only error that could potentially occur, right? That way we still check in parallel without violating the dependency requirement or causing any race conditions

src/tox/tox_get_outliers.F90
schroederaaron: can't we use do concurrent here to fill the perm? Or is the overhead more than the gain we would have?

schroederaaron (changes requested):

src/tox/data/integration/tox_data_integration_jsd.F90
qwerdenkerXD: 1. yes, the neighbors are not concurrent, because of the count. That is a race condition, indeed.
2. No, reduce doesn't work for arrays
3. atomic operations are meant for coarrays, so doesn't apply here

src/tox/tox_get_outliers.F90
qwerdenkerXD: We can use init_perm from f42_utils

Anyway, the allocations should use M_ALLOCATE as well, and some other loops can be parallelized, I think.

I don't remember why I stopped here

src/tox/tox_errors.F90
qwerdenkerXD: any(vec > x) creates a temporary for vec>x, that's why

src/tox/tox_loess.F90
schroederaaron: could use do concurrent here

schroederaaron: could use do concurrent here

src/tox/tox_trajectory_normalization.F90
schroederaaron: I don't think the outer loops are possible to be set to do concurrent. If the samples/factors are set to concurrent, the `tmp_series` array will cause a race condition if not set to local. So we would need a local copy of the array in each parallel execution. Also, the call with `ierr` would cause a race condition, as we then get multiple `normalize_variable_timeseries` calls that write the same error code at the same time. 

Still, i think we do not need do concurrent at all here as we are only copying elements, aren't we? Wouldn't a normal `tmp_series = trajectories(i_factor, i_sample, :)` be at least as efficient or maybe even more? Currently we tell the compiler "Copy all `i` elements in a loop into the new array, those operations are independent". But wouldn't it make more sense to directly tell the compiler that the full section of the array is going to be copied anyways?

src/tox/tox_trajectory_contribution_analysis.F90
schroederaaron: Same as in the normalization, do we need do concurrent here? Wouldn't a simple `tmp_dependent = trajectories(dependent_idx, random_sample, :)` be better here as every timepoint is copied anyways? Same goes for the `tmp_factor` above.

src/tox/tox_trajectory_contribution_analysis.F90
schroederaaron: I think we should keep it as it is. To parallelize the outer loop we could refactor it a bit and move the concurrent time point loop to the outer loop and then iterate over the permutations in the inner non-concurrent loop if i am seeing it correctly. But that would change our cache behaviour as we currently have a very cache friendly version with fast changing indices on continous strides, but if we switch the loops it would be less memory efficient. Not sure if that trade-off would be worth it, so i would keep it as it is.

src/tox/tox_trajectory_contribution_analysis.F90
qwerdenkerXD: `trajectories(dependent_idx, random_sample, :)` is not contiguous → temporary

src/f42/f42_utils.F90
VivianBass: Sorry I forgot to ask before but I already made a [comment](https://gitlab.rlp.net/aaron.schroeder/bachelor-thesis-aaron/-/work_items/2#note_941809) in the issue.

src/f42/f42_utils.F90
qwerdenkerXD: would not be standard compliant

src/tox/data/integration/tox_data_integration_jsd.F90
qwerdenkerXD: done

src/tox/tox_errors.F90
qwerdenkerXD: tox_errors doesn't do concurrent operations now

src/tox/tox_errors.F90
qwerdenkerXD: tox_errors doesn't do concurrent operations now

src/tox/tox_euclidean_distance.F90
qwerdenkerXD: corrected

src/tox/tox_normalization.F90
qwerdenkerXD: if you mean error codes, this is fixed. I don't see marked code lines here

src/tox/tox_get_outliers.F90
qwerdenkerXD: fixed

qwerdenkerXD: fixed

src/tox/tox_loess.F90
qwerdenkerXD: done

qwerdenkerXD: done

src/tox/tox_normalization.F90
qwerdenkerXD: I see the line now, is fixed

src/tox/tox_paralog_analysis.F90
qwerdenkerXD: not really. The helper will always return zero positions, as the `n_*` variables are guaranteed to be valid. And zero means internal error

src/tox/tox_paralog_analysis.F90
qwerdenkerXD: yes, but in the MR for merging the module, this was decided to be kept

src/tox/data/integration/tox_data_integration_jsd.F90
qwerdenkerXD: Done — verified in current code. The inner `i_neighbor` loop is now a plain sequential `do` (not `do concurrent`), since different neighbors can hit the same `(i_point, bin_idx)` count simultaneously. A comment in the code now explains why it can't be `do concurrent`.

src/tox/data/integration/tox_data_integration_jsd.F90
qwerdenkerXD: Done — verified in current code. `bin_width` now defaults to `1.0` when `shared_residual_range <= 0`, guarding against the division by zero.

src/tox/tox_normalization.F90
qwerdenkerXD: Done — verified in current code. Fixed via a plain sequential loop with a per-iteration `tmp_ierr` instead of writing the shared `ierr` inside `do concurrent`.

src/tox/tox_get_outliers.F90
qwerdenkerXD: Done — verified in current code, same fix as the other non-deterministic-`ierr` threads (sequential loop + `tmp_ierr`).

qwerdenkerXD: Done — verified in current code, same fix as the other non-deterministic-`ierr` threads (sequential loop + `tmp_ierr`).

qwerdenkerXD: Done in 0c323fd9. `init_perm` from `f42_utils` is now used instead of the manual loop, and the workspace allocations in `compute_family_scaling_helper` now go through the `M_ALLOCATE` macro instead of a manual `allocate(..., stat=istat)` + `check_io_stat`.

src/tox/tox_euclidean_distance.F90
qwerdenkerXD: Done — verified in current code. The C wrapper now documents `centroids` as `(n_elements × n_families)`, matching how `distance_to_centroid` actually treats it.

src/tox/tox_errors.F90
qwerdenkerXD: Done — verified in current code. `tox_errors` no longer uses `do concurrent` for this; it's a plain sequential loop now with a comment explaining why.

src/tox/tox_errors.F90
qwerdenkerXD: Done — verified in current code, same fix (plain sequential loop, no more `do concurrent`).

src/tox/tox_errors.F90
qwerdenkerXD: The immediate race (`do concurrent` over a shared `ierr`) is fixed — verified in current code, now a plain sequential loop. The broader point about switching some remaining status codes to status-code arrays is still open; tracking that separately as follow-up work rather than closing it here.

src/tox/tox_loess.F90
qwerdenkerXD: Done — verified in current code, `do concurrent` is used here now.

qwerdenkerXD: Done — verified in current code, `do concurrent` is used here now.

src/tox/tox_paralog_analysis.F90
qwerdenkerXD: Resolving as discussed — this was intentionally kept as-is per the decision in the module-merge MR. No code change needed.

src/tox/tox_paralog_analysis.F90
qwerdenkerXD: Resolving as discussed — same as above, kept as-is per the module-merge MR decision.

src/f42/f42_utils.F90
qwerdenkerXD: Done in 0c323fd9. `lower_bound_ge` has been removed; the call site in `compute_empirical_p_values` now uses `binary_search_insertion` (same lower-bound semantics, including the NaN handling), and `test/mod_test_empirical_pvalue.F90` was updated to call `binary_search_insertion` directly.

src/f42/f42_utils.F90
qwerdenkerXD: Done in 0c323fd9. Renamed the local `norm1`/`norm2` accumulators in `angle_between` to `norm1_sq`/`norm2_sq` so they no longer shadow the `norm2` intrinsic.

src/f42/f42_utils.F90
qwerdenkerXD: Done in 0c323fd9. `real_less`/`real_greater` now take `lhs, rhs` instead of the generic `a, b`.

rcpp/test/mod_test_which.R
qwerdenkerXD: Done in 0c323fd9. Removed the `which_c` C wrapper and the generated `tox_which` bindings in R (`rcpp/tensoromics_functions.R`/`.cpp`) and Python (`python/tensoromics_functions.py`), plus the now-obsolete `rcpp/test/mod_test_which.R` and `python/test/mod_test_which.py`. `which` stays Fortran-only as agreed.

test/mod_test_binary_search.F90
qwerdenkerXD: Done in 0c323fd9. The `TOL` constant was actually unused in this file, so it's been removed outright rather than just re-valued.

test/mod_test_clock_hand_angles.F90
qwerdenkerXD: Done in 0c323fd9 — `PI` is now imported from `f42_utils` instead of being redefined locally. Note: `TOL` is kept as a local `1e-12_real64` since `f42_utils` doesn't export a shared tolerance constant to import (only a machine-epsilon-based `EPS`, which is a different thing).

test/mod_test_trajectory_contribution_analysis.F90
qwerdenkerXD: Done in 0c323fd9. Replaced the overly tight `epsilon(1.0_real64)` tolerance with `1e-12_real64`, matching the convention used in sibling test files (e.g. `mod_test_clock_hand_angles.F90`, `mod_test_clustering.F90`).

qwerdenkerXD:
## Review thread checklist (updated)

Went through all 27 open threads, verified each against the current code, and fixed the clear-cut ones in `0c323fd9`. Individual replies posted on each thread below; summary here for convenience.

### ✅ Done (21)

- [x] [`tox_data_integration_jsd.F90:245`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3508917893) — histogram race fixed (sequential inner loop)
- [x] [`tox_data_integration_jsd.F90:237`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3508948912) — `bin_width` division-by-zero guarded
- [x] [`tox_normalization.F90:471`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3475213564) — non-deterministic `ierr` fixed
- [x] [`tox_get_outliers.F90:187`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3475246105) — non-deterministic `ierr` fixed
- [x] [`tox_get_outliers.F90:263`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3475252208) — non-deterministic `ierr` fixed
- [x] [`tox_get_outliers.F90:166`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3514472702) — now uses `init_perm`, and workspace allocations now go through `M_ALLOCATE`
- [x] [`tox_euclidean_distance.F90:178`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3475323903) — dimension doc corrected to `n_elements × n_families`
- [x] [`tox_errors.F90:408`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3475345187) — `do concurrent` race removed
- [x] [`tox_errors.F90:290`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3475350415) — `do concurrent` race removed
- [x] [`tox_errors.F90:370`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3475352483) — immediate race fixed; the broader "some status codes should become arrays" point is still open, tracked as follow-up (see below)
- [x] [`tox_loess.F90:356`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3520153177) — `do concurrent` now used
- [x] [`tox_loess.F90:366`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3520154564) — `do concurrent` now used
- [x] [`tox_paralog_analysis.F90:519`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3334439296) — resolved by discussion, kept as-is per module-merge MR decision
- [x] [`tox_paralog_analysis.F90:560`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3334447449) — same, resolved by discussion
- [x] [`f42_utils.F90:1490`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3334051295) — `lower_bound_ge` removed, replaced with `binary_search_insertion` at the call site and in tests
- [x] [`f42_utils.F90:398`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3333894899) — `norm1`/`norm2` locals renamed to `norm1_sq`/`norm2_sq`, no longer shadows the `norm2` intrinsic
- [x] [`f42_utils.F90:1055`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3333947397) — `real_less`/`real_greater` args renamed `a,b` → `lhs,rhs`
- [x] [`rcpp/test/mod_test_which.R:1`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3330055739) — `which_c` wrapper and the generated R/Python `tox_which` bindings removed; `which` stays Fortran-only
- [x] [`test/mod_test_binary_search.F90:13`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3334861548) — unused `TOL` artifact removed
- [x] [`test/mod_test_clock_hand_angles.F90:13`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3334872570) — `PI` now imported from `f42_utils`
- [x] [`test/mod_test_trajectory_contribution_analysis.F90:14`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3334987639) — `TOL` changed from `epsilon(1.0)` to the standard `1e-12`

### ❌ Still open (6)

- [ ] [`tox_paralog_analysis.F90:270`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3334425685) — disagreement never actually settled, needs a reviewer response
- [ ] [`f42_utils.F90:1443`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3334045867) — "empirical p-value" naming/statistics, blocked on Asis's answer on the linked GitLab issue
- [ ] [`tox_trajectory_normalization.F90:70`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3474529590) — confirmed real inconsistency: docstring says `(n_factors, n_timepoints)`, array is declared `(n_timepoints, n_factors)` — needs a decision + fix
- [ ] [`tox_trajectory_normalization.F90:139`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3474637211) — `do concurrent` grain-size / array-slice-copy question, deferred by consensus
- [ ] [`tox_trajectory_contribution_analysis.F90:259`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3474717086) — parallelize over permutations vs. time, cache-locality trade-off, deferred by consensus
- [ ] [`tox_trajectory_contribution_analysis.F90:178`](https://github.com/asishallab-group/Tensor-Omics/pull/135#discussion_r3474750119) — thread-local work array refactor, concrete plan sketched but deliberately deferred

The last three (plus the `tox_errors.F90:370` follow-up above) are good candidates to split into a dedicated follow-up issue for revisiting `do concurrent` granularity — all participants already agreed to defer them rather than block this PR.

qwerdenkerXD:
@VivianBass @schroederaaron This PR has grown quite large in its threads, thus I summariezed the done and open threads [above](https://github.com/asishallab-group/Tensor-Omics/pull/135#issuecomment-4979603755).

Please tell me which of the still open ones should move into a separate issue, which should be discarded and which applied here. We really need to come to an end here soon.

src/f42/f42_utils.F90
VivianBass: Asis has now clarified the intended interpretation. Please adjust the naming and documentation accordingly, so that the result is described as a scaled-distance quantile/effect-size measure rather than a p-value.

VivianBass:
> @VivianBass @schroederaaron This PR has grown quite large in its threads, thus I summariezed the done and open threads [above](https://github.com/asishallab-group/Tensor-Omics/pull/135#issuecomment-4979603755).
> 
> Please tell me which of the still open ones should move into a separate issue, which should be discarded and which applied here. We really need to come to an end here soon.

It’s a bit unfortunate that we can’t reply directly to comments. I’d suggest opening separate issues for the remaining open discussions so they don’t block the PR and we can bring this one to a close.

Maybe we could keep the re-naming one here if it’s not too much work, and open new issues for the remaining topics. If you think the naming discussion should also be moved to a separate issue, that’s perfectly fine as well.

VivianBass:
This PR feels very close to being wrapped up. From what I saw in the channel, you're going to tackle the naming in this PR, so once that's done it'll be approved on my end. Thanks so much for this great work, Franz!

PR #139, opened 2026-05-21
137 - Update guidelines

### LINKED ISSUE #137: Update Coding Guidelines
Our coding guidelines are somewhat outdated. This e.g. is true for our memory management, which we now have simplified greatly without loosing efficiently. Also recommendations about parallelization need to be clearly defined, i.e. use `do concurrent` everywhere where possible and let the compiler parallelize using vectorization or threads. We should mention the obligatory use of macros e.g. for automated error code setting memory allocation. 

Please update the guidelines according to our latest experiences. 

I think @VivianBass and @qwerdenkerXD should guide this issue using input from both @Specter1999 and @schroederaaron.

VivianBass:
The guidelines were built from scratch based on our previous document and the points suggested by the team. 

This PR closes issue #137 and #150


misc/Fortran_Coding_Guides.tex
```diff
@@ -28,1778 +41,1018 @@
...
 \begin{itemize}
-\item Use \texttt{!\$omp parallel do simd schedule(static)} to enable both thread and instruction-level parallelism.
-\item Use compiler-specific alignment and vectorization flags in \texttt{fpm.toml}.
-\item Dynamically set alignment to 32 or 64 via a shell script based on CPU capabilities.
-\item Use \texttt{padded\_size()} to round up all dimensions to nearest alignment multiple.
-\item Allocate arrays explicitly using padded sizes to ensure SIMD-safe alignment.
-\item Maintain a \texttt{config} module to centralize the alignment parameter.
+  \item \texttt{!>} — Starts the description of a module, subroutine, or function.
+  \item \texttt{!!} — Continues a description on the next line (trail comment after a declaration).
+  \item \texttt{!|} — Documents an argument; place this comment \textbf{on the line immediately after} the argument declaration, not before it.
```
qwerdenkerXD: actually `!|` needs to be placed before the declaration, like
```fortran
!| bla is cool
integer, intent(int) :: bla
```

It is only different in the preceding
```fortran
!> bla is
!| very cool
subroutine bla
```
This is just for prettiness

misc/Fortran_Coding_Guides.tex
```diff
@@ -28,1778 +41,1018 @@
...
+  M_DEFAULT_VAL(degree, actual_degree, 2_int32)
+\end{lstlisting}
 
-\section{Fortran Array Notation Cheat-Sheet}
+\textbf{Never} use optional arguments in C wrappers; \texttt{bind(C)} is incompatible with them.
 
-This section summarizes the three common ways to declare array arguments in Fortran procedures, with a focus on their use, safety, and suitability in modern code.
+% =============================================================================
+\section{Parallelisation: \texttt{do concurrent}}
+% =============================================================================
```
qwerdenkerXD, line 478: concurrent needs a part that explains the locality specs. Each variable used in a concurrent loop needs a locality spec, otherwise the compiler assumes it.
- `shared(...vars)`: vars are shared across iterations
- `local(...vars)`: each iteration gets its own local variable (local temporaries)
- `reduce(operator: ....vars)`: used for scalars with accumulating operator, like adding values
   ```fortran
   means_sum = 0.0_real64
   do concurrent (i_gene = 1:n_genes) shared(gene_means) reduce(+:means_sum)
       means_sum = means_sum + gene_means(i_gene)
   end do
   ```
   Supported operators: `iand`, `ieor`, `ior`, `max`, `min`, `+`, `*`, `.and.`, `.or.`, `.eqv.`, `.neqv.`

misc/Fortran_Coding_Guides.tex
qwerdenkerXD, line 1: Maybe add a reference to the fortran standard draft as lookup.
https://j3-fortran.org/doc/year/26/26-007.pdf

qwerdenkerXD, line 1: I think a convention for indentation would be good. I propose 4 whitespaces.
This might also help: `fprettify -i 4 -l 1000 <path-to-file>`

qwerdenkerXD, line 1: Also, only one dummy argument declaration per line, so each can get an own Ford comment. For locals this is not needed, but they should have meaningful names though, as all variable names in general

misc/Fortran_Coding_Guides.tex
```diff
@@ -28,1778 +41,1018 @@
...
+\end{lstlisting}
 
-\section{Critical Practices in TOX Interfacing and Control Flow}
+If these kinds do not match, the module fails to compile with a descriptive error, preventing silent ABI (Application Binary Interface) mismatch bugs. The ABI is the low-level contract between compiled code: it specifies how function arguments are laid out in memory, what sizes the types occupy, and how values are returned. When the Fortran kind for \texttt{real64} (8 bytes) does not match the C kind for \texttt{c\_double} on a given platform, data passed through a C wrapper would be misinterpreted silently, producing incorrect results without any runtime error.
 
-This section consolidates key implementation standards for TOX and F42 compliance: prohibition of \texttt{GOTO} and manual memory cleanup.
+% =============================================================================
+\section{C Wrapper Conventions}
+\label{sec:cwrappers}
+% =============================================================================
```
qwerdenkerXD, line 728: should also include logical handling

And the arguments should keep the Ford doc from their original routine

qwerdenkerXD (changes requested):

misc/Fortran_Coding_Guides.tex
qwerdenkerXD: But for `!!` it comes after
```fortran
integer, intent(int) :: bla
    !! bla is cool
```

VivianBass: Done

misc/Fortran_Coding_Guides.tex
VivianBass: Done

misc/Fortran_Coding_Guides.tex
VivianBass: Done

VivianBass: Done

VivianBass: Done

misc/Fortran_Coding_Guides.tex
VivianBass: Done

misc/Fortran_Coding_Guides.tex
```diff
@@ -28,1778 +41,1226 @@
...
+#define CM_LOESS_SPAN_DEFAULT 0.7_real64
+#define CM_LOESS_DEGREE_DEFAULT 2_int32
+\end{lstlisting}
 
-\begin{itemize}
-    \item Scalars (e.g., \texttt{integer}, \texttt{real}) are passed by reference.
-    \item Arrays (e.g., \texttt{integer, dimension(:)}, \texttt{real, dimension(:,:)}) are passed by reference.
-    \item Derived types are passed by reference.
-\end{itemize}
+The \texttt{CM\_} prefix (``constant macro'') clearly distinguishes them from regular variables.
```
qwerdenkerXD: maybe I should explain that:
When initially introducing macros, I defined them inside `macros.h` with the `M_` prefix and all uppercase, so they can be identified as macros directly when reading through code.
Then there was a case I needed a macro to not repeat myself that was only related the one single module. Defining it inside `macros.h` would've been a little bit out of purpose, as the ones defined there are reusable across our codebase. This is why I chose the `CM_` prefix for macros defined directly inside a file. It actually stands for `Custom Macro`.
It is used very often for constants, so I can use them in Ford comments as well. For example, when writing the Ford doc for a mode integer variable, I can simply create the table with a mapping there, while using the macros for the values. Then, if we ever introduce more modes etc that may change the mode values, everything is still consistent. This is also why we have the macros then and them also being assigned to the Fortran `parameter` constants. Ford doesn't resolve these constants in documentation, *but* it does preprocessing.
Another use case I remember is to validate modes, like `CM_VALIDATE_MODE_<whatever>(ARGPOS) ...` that expands to something like `call validate_in_range_int(mode_<whatever>, ierr, min=CM_FIRST_MODE, max=CM_LAST_MODE, ARGPOS)`. The argpos thing is something new in #135 that adds argument positions to error codes.

Such `CM_` macros will usually be defined at the top where the module constants are defined as well, so all constant things lie in one place. If anything changes in their values, the related macros can be updated directly.

misc/Fortran_Coding_Guides.tex
```diff
@@ -28,1778 +41,1226 @@
...
+\subsection{Available Macros}
+
+\begin{tabular}{ll}
+\toprule
+Macro & Purpose \\
+\midrule
+\texttt{M\_USE\_NULL\_VALIDATION} & Imports modules needed for \texttt{c\_loc}/null checks in C wrappers \\
+\texttt{M\_CHECK\_IERR\_NON\_NULL} & Silently returns if \texttt{ierr} pointer is null (always first check) \\
+\texttt{M\_CHECK\_NON\_NULL(ARG)} & Sets \texttt{ERR\_POINTER\_NULL} and returns if \texttt{ARG} pointer is null \\
+\texttt{M\_DEFAULT\_VAL(OPT,LOC,DEF)} & Assigns \texttt{OPT} to \texttt{LOC} if present, else assigns \texttt{DEF} \\
```
schroederaaron: I would rephrase this to "Sets `ierr` to ... " as it currently sounds like the ERR_POINTER_NULL or else is a variable or similar in my opinion

misc/Fortran_Coding_Guides.tex
```diff
@@ -28,1778 +41,1226 @@
...
-!$omp end parallel do simd
-\end{verbatim}
+\subsection{Key Points}
 
-In this example:
 \begin{itemize}
-  \item \texttt{i} is the loop counter and must be private.
-  \item \texttt{tmp} is a per-iteration local value and must be private to avoid race conditions.
-  \item \texttt{sum} is a global accumulator and must be declared as a reduction variable with the \texttt{+} operator.
+  \item 2D arrays passed to Fortran must be \textbf{Fortran-contiguous} (\texttt{np.asfortranarray}); 1D arrays can be C-contiguous.
```
schroederaaron: I think this statement can be improved. First of all, it applies for all arrays with dimension > 1, not only 2D. Also, arrays can be created with the `order='F'` argument in numpy which should force the fortran ordering. And finally, np.asfortranarray creates a 1D copy of the array that has the correct memory layout for fortran. It is better to create an array directly with `order='F'` to avoid that copy.

That how i understood it [here](https://numpy.org/devdocs/reference/generated/numpy.asfortranarray.html)

schroederaaron (changes requested):

misc/Fortran_Coding_Guides.tex
qwerdenkerXD: it will only copy if not already fcontiguous, see last example in the docs:
> Now, starting with a Fortran-contiguous array:
>
>x = np.ones((2, 3), order='F')
x.flags['F_CONTIGUOUS']
True
>
>Then, calling asfortranarray returns the same object:
>
>y = np.asfortranarray(x)
x is y
True

schroederaaron: Yes that is true. My point was to add the `order='F'` part to make sure that we avoid copies as good as possible. It is still important to add the `np.asfortranarray` especially for array created by a user to be sure that we have the right layout, but for code written by us we can avoid possible copies by using the order flag

misc/Fortran_Coding_Guides.tex
```diff
@@ -28,1778 +41,1226 @@
...
-do i = 1, n
-  tmp = compute(i)
-  sum = sum + tmp
-end do
-!$omp end parallel do simd
-\end{verbatim}
+\subsection{Key Points}
 
-In this example:
 \begin{itemize}
```
schroederaaron, line 1152: I would add that all functions that return more than one value have to return a dictionary.

Specter1999:
In the meeting part that I mention:

https://github.com/asishallab-group/Tensor-Omics/pull/143#discussion_r3356292500

Specter1999:
Section 3.2 explicitly mandates one dummy argument declaration per line to ensure proper FORD documentation.

However, the example snippets under Section 6 (specifically lines like integer(int32), intent(in) :: n_genes, n_replicates) accidentally group the dummy arguments on a single line.

misc/Fortran_Coding_Guides.tex
VivianBass: I've reworked the paragraph, I hope it now aligns better. I also added the argpos thing to the guidelines.

misc/Fortran_Coding_Guides.tex
VivianBass: Done

misc/Fortran_Coding_Guides.tex
VivianBass: Done

misc/Fortran_Coding_Guides.tex
VivianBass: Done

VivianBass:
> Section 3.2 explicitly mandates one dummy argument declaration per line to ensure proper FORD documentation.
> 
> However, the example snippets under Section 6 (specifically lines like integer(int32), intent(in) :: n_genes, n_replicates) accidentally group the dummy arguments on a single line.

Done.

qwerdenkerXD (approved):

Issue #136, opened 2026-05-15
Implementation of Generalized Operators for Ensemble Growth and Clustering


Specter1999:
# Generalized functions for "operators"

### 1. Coarse graining or better `increase_ensemble`:
* **Arguments:** set of vectors (ensemble) probably as a matrix or a logical selecting column vectors (indices) of "ambient_space", ambient space (the superset of all vectors), further arguments
    * There can be different functions that we implement here like density based, energy based, angular based...
* **Return value (output):** A grown ensemble or the same if growth was rejected

### 2. Feature
* **Arguments:** An ensemble and optional further arguments
* **Returns:** A real vector representing the ensemble without scale
    * Can be: Density, Angular Variance, First Singular Vector, Manifold, Free Energy, Rank, ....

### 3. Normalization
* As of now, I'm unsure whether we need this or whether this will be an integrated part of the `Feature` function family
* **Arguments:** An ensemble (size), the feature,
* **Returns:** The feature normalized to the size, i.e. number of member vectors of the Ensemble

---

# Clustering algorithm implementation:

1. Starts with an initial **Ensemble**
2. Gets its **feature**
3. **Grows** the ensemble
4. Gets the **feature** for the grown ensemble
5. Checks the **difference** between Feature $F_{i+1}$ and $F_{i}$, potentially using the first derivative
6. **Outputs** the iterations results:
    * **Ensemble and their features**
    * **Flow of Feature differences**

asishallab:
# Meeting notes from May 28th 2026

* We will need a function to get starting ensembles. In our current approach based on density labels this would be the top $t$ (parameter) most dense vectors. 
  * Bear in mind we try to achieve two things at this moment:
    * A general and "interface like" definition of how to do physics inspired clustering (renormalization and coarse graining)
    * A first (test) implementation based on density 
* For what a fixed point is see the below sketch, please
* We will need a function to define a "neighborhood", so a subset of the ambient space that contains _candidate_ vectors to be added to an existing ensemble. This potentially can be part of the "grow" function

<img width="2450" height="1379" alt="Image" src="https://github.com/user-attachments/assets/7ca64078-afc5-403d-b6cd-4d0f206f267f" />

Specter1999:
**1. Coarse graining or better increase_ensemble:**

[...]

PR #143, opened 2026-05-28
Shape truthful clustering

### LINKED ISSUE #133: Shape-Truthful-Clustering: Get started
This issue is to get the new project of shape truthful clustering started.

# Coding Guides

Please consult and follow strictly our Fortran Coding Guides in the `misc` directory. Especially (almost) all functions should be `pure` and accept only preallocated arrays, also for intermediate data storage and return values. Write `ISO_C_BINDING` API functions for each function you implement in Fortran. _Later_ implement the Python and Rcpp & R wrappers for those ISO_C API functions. 

# Setup

- create a new branch called `shape-truthful-clustering` maybe dub the method "shatter"
- Transfer (don't invest too much time) our slide content into a section in our `misc/Tensor_Omics_Methods` Latex section
  - Text should be copy-pasted
  - Manually drawn figures can be screenshot and used as figures 

# Data structures

Define reasonable Fortran Structure of Arrays, preallocated array based, data structures for our clustering method.

Input will be a two dimensional array of vectors, where columns are the vectors and rows represent dimensions of the multidimensional vector space. For now expect real arrays.

We need another one or two dimensional (assumed shape array) for the vector labels, which can be density, cone angular variance, energy, main axis of variance, etc. values. The column indices match (one-to-one relation) column indices in the input data vectors, so we can identify vector labels efficiently. Labels can be any vector (single entry, i.e. scalar or multi-dimensional feature vector). 

## Label calculation

Define an interface for label calculation that accepts the input data vectors (two dim. array) and a preallocated label array. Offer several default label calculation functions. We start with density labels. Make sure the interface will work with assumed shape arrays for labels, so any dimension, i.e. number of rows, can be provided.

Most likely an interface will not solve the issue to offer different functions to calculate vector or scalar (1d vector) labels. In that case use prefix naming based grouping and call all different label functions:
`calculate_labels_as_[...]` e.g. `calculate_labels_as_density`

#### Calculate a label-sphere radius

Implement a function that calculates a good radius for the input data vectors. First iteration algorithm:
- get the mean vector
- calculate all distance of the mean vector to all other vectors
- take the `mean_to_other_vecs_dist_quant` parameter with default set to `0.15` as the radius

Alternative implementation to test later:
- for all data vectors use a kNN approach with a meaningful min k, e.g. max(no. of data vectors divided by 1000, 30) 
- for each vector get the distances to its k nearest neigbor
- collect those distances and use the (parameter) x quantile as radius

#### Construct a K-D-Tree of the input data vectors

Use our existing methods to construct a K-D-Tree of our input data vectors and pass it to the below function.

#### Density Label the data vectors
- implement a function to calculate the density labels for data vectors
  - it should accept a parameter `r` (see Calculate a label-sphere radius), the above k-d-tree, and a preallocated label array, and of course the data vectors themselves
- For each input data vector _Vi_:
  - Using the k-d-tree get all vectors in _Vi_ label sphere (radius)
  - calculate the density as number of vectors per sphere

Specter1999:
This is my first Fortran module; therefore, there might be many errors.  

**Methods** 
[tox_shatter_cluster_data.F90](https://github.com/asishallab-group/Tensor-Omics/blob/shape-truthful-clustering/src/tox_shatter_cluster_data.F90)

**Tests for tox_shatter_cluster_data.F90**
[mod_test_shatter_cluster_data.f90](https://github.com/asishallab-group/Tensor-Omics/blob/shape-truthful-clustering/test/mod_test_shatter_cluster_data.f90)

**Just added call and use for mod_test_shatter_cluster_data.f90**
[run_tests.f90](https://github.com/asishallab-group/Tensor-Omics/blob/shape-truthful-clustering/test/run_tests.f90)

**Method definitions**
[Shatter_Clustering.tex](https://github.com/asishallab-group/Tensor-Omics/blob/shape-truthful-clustering/misc/Tensor_Omics_Methods/Shatter_Clustering.tex)
[Tensor_Omics_Methods.pdf](https://github.com/asishallab-group/Tensor-Omics/blob/shape-truthful-clustering/misc/Tensor_Omics_Methods.pdf)

Closes #133 #136 #169


src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,346 @@
...
+		real( real64 ), allocatable :: tmp_mean_vec(:)
+		real( real64 ), allocatable :: tmp_distances(:)
+		integer( int32 ), allocatable :: tmp_perm(:)
+
+		call set_ok( ierr )
+
+		! Perform structural allocations using the project safety macro
+		M_ALLOCATE( tmp_mean_vec( n_dimensions ) )
+		M_ALLOCATE( tmp_distances( n_vectors ) )
+		M_ALLOCATE( tmp_perm( n_vectors ) )
```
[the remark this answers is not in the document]
qwerdenkerXD: correct, please use the validation routines from `tox_errors`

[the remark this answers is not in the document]
Specter1999: We also have an updated Fortran coding guide then, right? Because in the _alloc it does not check and happens in the second layer. And since we check in the _alloc layer, do we still need to check in the second layer?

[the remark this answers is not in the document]
qwerdenkerXD: non-helpers do validations and call the helper routine. This also means to assign default values in the helper, not inside the non-helpers.

Also, I think we should expect quantiles directly to be in range 1-100. On the other hand, it's really annoying that we have this 1-100 range at all. In calculations, we usually get fractions, just to multiply by 100 then, which is divided by 100 again in the percentile calculation.
We could pick that up in the meeting.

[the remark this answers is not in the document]
Specter1999: Alright, then I am also adding checks inside the _alloc layer. 

Yeah, I also felt strange when I was doing the code. In the issue, Prof. Hallab mentions that function expect values between 0 and 1, but when I do call it, it is already dividing 100, so I need to do multiplication before calling. Better to pick up in the meeting.

test/mod_test_shatter_cluster_data.f90
```diff
@@ -0,0 +1,153 @@
...
+                    print *, " ", trim(test_names(i_name)), " passed."
+                    found = .true.
+                    exit
+                end if
+            end do
+            if (.not. found) then
+                print *, "Unknown test: ", trim(test_names(i_name))
+            end if
+        end do
+    end subroutine run_named_tests_shatter_cluster_data
```
qwerdenkerXD: remove both routines. The test suite defines one for all

test/mod_test_shatter_cluster_data.f90
```diff
@@ -0,0 +1,153 @@
...
+            vectors(:, i_setup) = [real(i_setup, real64), real(i_setup, real64)]
+        end do
+
+        ! 2. Act
+        call calculate_label_sphere_radius(vectors, n_dims, n_vecs, &
+                                           tmp_mean_vec, tmp_distances, tmp_perm, &
+                                           radius, 0.50_real64, ierr)
+
+        ! 3. Assert
+        call assert_true(is_ok(ierr), "test_sphere_radius_basic: ierr success control check")
```
qwerdenkerXD: I recommend `assert_equal_int(ierr, ERR_OK, ...` for error code checks (with respective expected error code). If the assertion fails, its assertion message shows received and expected value. Helps for debugging

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,346 @@
...
+
+            do k_vec = 1, n_vectors
+                target_idx = kd_indices( k_vec )
+
+                ! Protect parallel core against array indexing overflows
+                if ( target_idx < 1 .or. target_idx > n_vectors ) then
+                    ! Note: internal loop runtime tracking via safe shared variables
+                    label_values(1, i_vec) = -1.0_real64 
+                    cycle
+                end if
```
qwerdenkerXD: if I'm not mistaken, this place is the only one introducing a sentinel. 
I think it's better to include `kd_indices` into input validation, as it is the user's fault if the indices are out of bounds.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,346 @@
...
+
+                call euclidean_distance( vectors(:, i_vec), vectors(:, target_idx), n_dimensions, calculated_dist )
+                if ( calculated_dist <= r ) then
+                    match_count = match_count + 1
+                end if
+            end do
+
+            label_values(1, i_vec) = real( match_count, real64 )
```
qwerdenkerXD: Why is `label_values` a matrix if you always set the column's first index only?

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,346 @@
...
+        call validate_dimension_size( n_dimensions, ierr )
+        call validate_dimension_size( n_vectors, ierr )
+        if ( is_err( ierr ) ) return
+        
+        ! Delegate straight down to your core indexer
+        call build_kd_index( vectors, n_dimensions, n_vectors, kd_indices, dimension_order, &
+                             tmp_workspace, tmp_value_buffer, tmp_permutation, &
+                             tmp_left_stack, tmp_right_stack, tmp_recursion_stack, ierr )
+
+    end subroutine compute_label_kd_tree
```
qwerdenkerXD: I don't think we need these routines. Just document in the Ford comments for kd indices where this arg comes from (Ford supports referencing code elements like `[[mod_name(module):sub_name(subroutine)]]`).

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,346 @@
...
+			call euclidean_distance( mean_vec, vectors( :, i_vec ), n_dimensions, distances( i_vec ) )
+		end do
+
+		call sort_real_heapsort( distances, perm )
+
+		percentile = quantile_frac * 100.0_real64
+
+		call calc_percentile( distances, perm, percentile, radius, ierr )
+
+	end subroutine calculate_label_sphere_radius_helper
```
qwerdenkerXD: The name of this routine (and its non-helper variants) is a little bit confusing. It calculates the `label sphere radius`, but has nothing to do with the labels you've calculated.

test/mod_test_shatter_cluster_data.f90
```diff
@@ -0,0 +1,153 @@
...
+        
+        ! Members 1 to 6 should find exactly 6 neighbors (including itself)
+        do i_chk = 1, 6
+            call assert_equal_real(label_values(1, i_chk), 6.0_real64, 1.0e-12_real64, "Dense cluster tracking verification")
+        end do
+        
+        ! Members 7 to 10 should find exactly 1 neighbor (only itself)
+        do i_chk = 7, 10
+            call assert_equal_real(label_values(1, i_chk), 1.0_real64, 1.0e-12_real64, "Isolated point density verification")
+        end do
```
qwerdenkerXD, line 149: this is zero tolerance. The labels are integer-like in the implementation

qwerdenkerXD (changes requested):

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,255 @@
...
+        ! Validate dimension sizes using core tox modules
+        call validate_dimension_size(n_dimensions, ierr)
+        call validate_dimension_size(n_vectors, ierr)
+        if (is_err(ierr)) return
+
+        ! Validate user index boundaries upfront to prevent running core checks
+        if (any(kd_indices < 1) .or. any(kd_indices > n_vectors)) then
+            call set_err(ierr, ERR_INVALID_INPUT)
+            return
+        end if
```
qwerdenkerXD: use validate_all_in_range_int

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,255 @@
...
+
+        ! Validate metric boundaries are non-negative
+        call validate_in_range_real(r, ierr, min=0.0_real64, max=huge(1.0_real64))
+        if (is_err(ierr)) return
+
+        ! Vector size verification matching n_vectors
+        if (size(label_densities) /= n_vectors) then
+            call set_err(ierr, ERR_DIM_MISMATCH)
+            return
+        end if
```
qwerdenkerXD: declare it as of size `n_vectors`, not `:`

qwerdenkerXD (changes requested):

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+
+        ! Boundary validation check for index sequence
+        call validate_all_in_range_int(kd_indices, n_vectors, ierr, min=1_int32, max=n_vectors)
+        if (is_err(ierr)) return
+
+        ! Validate search metrics are non-negative
+        call validate_in_range_real(r, ierr, min=0.0_real64, max=huge(1.0_real64))
+        if (is_err(ierr)) return
+
+        ! 🚀 FIXED CALL ALIGNMENT: Arguments here now perfectly match the helper signature below
```
qwerdenkerXD: be cautious with generated comments:
1. nobody cares about the information that something has been fixed at some time
2. only ascii

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+        real(real64), intent(in) :: r
+        !! Label-sphere search radius
+        integer(int32), intent(in) :: dimension_order(n_dimensions)
+        !! Dimension split order array tracking the tree structure
+        integer(int32), intent(in) :: kd_indices(n_vectors)
+        !! KD-tree index sequence array computed via build_kd_index
+        real(real64), intent(out) :: label_densities(n_vectors)
+        !! Output density tracker matching individual vector slots
+        integer(int32), intent(out) :: ierr
+        !! Error code tracker (always final parameter)
```
qwerdenkerXD: I assume the `always final parameter` has been appended by AI because it is the last one in the argument list? Please remove that, it is confusing. Maybe just use `!! Error code` as comment

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+
+        call set_ok(ierr) ! ierr = ERR_OK = 0
```
qwerdenkerXD: Separation of concerns: We have this routine so we don't need to care about the underlying code. Commenting the assigned value is confusing if we ever rename error codes or their values/encoding

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+        integer(int32) :: match_count
+        integer(int32) :: stack_top
+        integer(int32) :: left_idx
+        integer(int32) :: right_idx
+        integer(int32) :: mid_idx
+        integer(int32) :: current_dim
+        integer(int32) :: current_depth
+        integer(int32) :: node_point_idx
+        real(real64) :: dist
+        real(real64) :: axis_delta
```
qwerdenkerXD: for locals please don't use one line per variable. This bloats the file

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+
+        ! Mandated tmp_ prefix applied to this loop-local work array
+        integer(int32) :: tmp_stack(3, 64)
```
qwerdenkerXD: 1. unnecessary comment
2. no allocation inside the helper. Take it as input, while _alloc allocates it

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+
+                if (axis_delta + r >= 0.0_real64) then
+                    if (mid_idx + 1 <= right_idx) then
+                        stack_top = stack_top + 1
+                        tmp_stack(1, stack_top) = mid_idx + 1
+                        tmp_stack(2, stack_top) = right_idx
+                        tmp_stack(3, stack_top) = current_depth + 1
+                    end if
+                end if
+            end do
```
qwerdenkerXD: what are you doing here?

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+        call validate_dimension_size(n_dimensions, ierr)
+        call validate_dimension_size(n_vectors, ierr)
+        if (is_err(ierr)) return
+
+        ! Boundary validation check for index sequence
+        call validate_all_in_range_int(kd_indices, n_vectors, ierr, min=1_int32, max=n_vectors)
+        if (is_err(ierr)) return
+
+        ! Validate search metrics are non-negative
+        call validate_in_range_real(r, ierr, min=0.0_real64, max=huge(1.0_real64))
```
qwerdenkerXD: max is optional, it is fine if you only pass min

src/tox_shatter_cluster_data.F90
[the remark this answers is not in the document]
qwerdenkerXD: did you talk with Asis about this? I remember this question was postponed to your personal meeting

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+
+        call set_ok(ierr)
+
+        ! Validate all input dimensions first
+        call validate_dimension_size(n_dimensions, ierr)
+        call validate_dimension_size(n_vectors, ierr)
+        if (is_err(ierr)) return
+
+        ! Assign default value using mandated preprocessor macro
+        M_DEFAULT_VAL(mean_to_other_vecs_dist_quant, actual_quant, 0.15_real64)
```
qwerdenkerXD: 1. assign default values inside the _helper
2. instead of assigning here, validate it is in range. All validation routines allow optional arguments, so passing `mean_to_other_vecs_dist_quant` directly to `validate_in_range_real` works

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+        real(real64), allocatable :: tmp_mean_vec(:)
+        real(real64), allocatable :: tmp_distances(:)
+        integer(int32), allocatable :: tmp_perm(:)
+
+        call set_ok(ierr)
+
+        ! Validate all input dimensions first
+        call validate_dimension_size(n_dimensions, ierr)
+        call validate_dimension_size(n_vectors, ierr)
+        if (is_err(ierr)) return
```
qwerdenkerXD: in validations also validate vectors to ensure they don't include NaN. The validation routine for real values automatically validates for NaN and infinity, thus `validate_all_in_range_real(vectors, size(vectors, kind=int32), ierr)` works.

Also, to avoid redundant validations then, call the _helper function instead of `calculate_density_radius`

qwerdenkerXD (changes requested):

Specter1999:
I accidentally pushed from the wrong directory, which is why there are so many AI-generated comments. Sorry for that mistake.

I will discuss this with him this week. Additionally, I am planning to discuss the _alloc layer, validations, and the Fortran coding guide. I believe one of them is wrong. For now, I did it as the Fortran coding guide suggested. After the meeting I will adapt accordingly.

src/tox_shatter_cluster_data.F90
Specter1999: Regarding the part you asked for:

- Our goal is to calculate the local point density for every vector within a given label-sphere radius (r).

- To begin with, instead of using recursive function calls, this implementation utilizes a preallocated 3D array (tmp_stack) to track the implicit K-D tree nodes during traversal. The code enters a loop that runs as long as there are regions left on the stack.

- For each node, the algorithm calculates the spatial distance along the current split axis (axis_delta). It evaluates the search radius (r) against this threshold to aggressively skip entire left or right sub-branches that cannot mathematically contain any qualifying neighbors.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,312 @@
...
+
+        call set_ok(ierr)
+
+        ! Validate dimension sizes, metrics, and sequences across inputs
+        call validate_dimension_size(n_dimensions, ierr)
+        call validate_dimension_size(n_vectors, ierr)
+        call validate_all_in_range_real(vectors, size(vectors, kind=int32), ierr)
+        if (is_err(ierr)) return
+
+        call validate_all_in_range_int(kd_indices, n_vectors, ierr, min=1_int32, max=n_vectors)
```
VivianBass: I think we should also validate dimension_order, right?

src/tox_shatter_cluster_data.F90
VivianBass: I think this behavior should live in f42_kd_tree as a subroutine like kd_range_count (or a better name), since it is a generic tree operation, not something specific to clustering.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,312 @@
...
+        real(real64), intent(out) :: label_densities(n_vectors)
+        !! Output density tracker matching individual vector slots
+        integer(int32), intent(out) :: ierr
+        !! Error code
+
+        integer(int32), allocatable :: tmp_stack(:, :, :)
+
+        call set_ok(ierr)
+
+        M_ALLOCATE(tmp_stack(3, 64, n_vectors))
```
VivianBass: I think the documentation should mention the 64 bound

VivianBass: Also, according to the guidelines we should define the numeric constants at the beggining, maybe something like:

`#define CM_KD_TRAVERSAL_STACK_DEPTH 64`

src/tox_shatter_cluster_data.F90
qwerdenkerXD: I'm just curious because this repeats nearly all of the code of the kd tree building. Shouldn't you use the `kd_indices` for traversal?

qwerdenkerXD:
@Specter1999 regarding the _alloc layer: You did not validate all input arguments. In this case it might work to simply do only some validations in the alloc routine, but sometimes these include O(n) validations, which would be duplicate in the non-alloc call.

This is why in such cases the _alloc routine calls the helper directly. If you want to call the non-alloc variant here, go for it, but you still need to add all validations (if you haven't already)

src/tox_shatter_cluster_data.F90
VivianBass: Another thing: the do while loop inside do concurrent is problematic: although the iterations are technically independent (each i_vec operates on its own tmp_stack(:,:,i_vec) slice), gfortran cannot vectorize or parallelize a body with variable-length do while loops, so it will silently fall back to serial execution, making the do concurrent useless. Extracting it into a kd_range_count subroutine in f42_kd_tree would give the compiler a clean call per iteration that it can actually parallelize. What do you think @qwerdenkerXD ?

qwerdenkerXD: I'm not sure if that makes a difference. But anyway, as far as I understand the traversal should simply happen by iterating through the kd indices and with that index retrieving the point. It should not need the nearly complete reimplementation of the kd tree construction.

@schroederaaron you have implemented kd trees. `get_kd_point` was meant for traversal, was it? Maybe give a short snippet on how to traverse kd trees.

VivianBass: Wouldn't iterating through kd_indices be a brute force scan? I think Salih needs a subroutine that given a radius returns the count of points or the points inside that radius, I don't think that can be done with anything that lives now in our kd_tree.

VivianBass:
The tests cover the happy path but are missing several important cases:

- No validation tests: `r < 0`, `quantile > 1.0` or `< 0.0`, and out-of-range `kd_indices`
  should all return a non-OK `ierr`. The validation layer is completely untested.

- `test_density_radius_basic` only checks that the radius falls in [0, 10] — it does not
  verify the actual value. With `quantile=0.50` on a linear diagonal dataset the expected
  radius is analytically known and should be asserted exactly (or within a small tolerance).

- The default quantile (0.15) is never exercised. There should be a test that omits the
  optional argument and verifies the result matches what 0.15 would produce.

- `test_density_labels_basic` only tests one radius value (0.5). A second test with a
  smaller radius (e.g. 0.05, where only directly adjacent points qualify) would confirm
  that the pruning logic is working correctly and not over-counting.

- Edge cases missing: `n_vectors = 1` (a single point must have density 1.0), and all
  vectors identical (every vector should see all `n_vectors` within any r > 0).

- Neither `calculate_density_radius_alloc` nor `calculate_labels_as_density_alloc` is
  tested — the allocating wrappers are untouched.

- The assert message in the density loop is the same string for all 10 vectors
  (`"Dense cluster tracking verification"`). If one fails it is impossible to know which
  vector caused it. Each assert should include the index.

src/tox_shatter_cluster_data.F90
Specter1999: I mainly agree with Vivian. This is mostly tree operation, and moving there will be better. About do concurrent loops, I think it will be better to have them, as Vivian suggested, to tackle the bottleneck problem, but I am not so sure. 

To clarify the traversal math: we can't just loop flatly over kd_indices or use get_kd_point alone, because that turns the query into a brute-force linear scan across all data points. To actually get better performance benefit of the KD-tree during a radius search, the algorithm has to evaluate the split axes and prune branches we don't need to look at.

schroederaaron: Yes i have implemented the kd trees back in the very beginning of the project. If I remember correctly the get_kd_point was meant for traversal to find a specific point. I can not surely say how exactly it is implemented, i would need to look that up.

Specter1999:
> @Specter1999 regarding the _alloc layer: You did not validate all input arguments. In this case it might work to simply do only some validations in the alloc routine, but sometimes these include O(n) validations, which would be duplicate in the non-alloc call.
> 
> This is why in such cases the _alloc routine calls the helper directly. If you want to call the non-alloc variant here, go for it, but you still need to add all validations (if you haven't already)

Yes, I am aware of that. Like I mentioned after pushing, I will adapt the code after discussing this part.

src/tox_shatter_cluster_data.F90
asishallab: Thank you @VivianBass, @qwerdenkerXD, @schroederaaron, and @Specter1999 for the lively discussion!

I agree with Vivian to extend the functionality of our K-D-Tree module to include a function / subroutine `vicinity_vectors` (or something similar) that returns a `logical` mask to the ambient vectors (the vectors used as input for the K-D-Tree creation). Such a mask will serve several purposes:
* Get all data points (vectors) that are in the vicinity (radius) of a query vector (function argument to `vicinity_vectors`
* As it is a mask it can be used to count those, too, with Fortran's `COUNT(mask)` function
* Also, as it is a `logical` mask we avoid copying the ambient space input vectors.

Actually, I believe we should check whether our K-D-Tree implementation makes maximum use of `logical` masks in order to be very careful with our memory - and compute time - usage. Note that copying also takes compute time..

Thank you and Cheers!

src/tox_shatter_cluster_data.F90
Specter1999: @VivianBass Does it also go for value 3 or only for 64?

VivianBass: yeah all constnts that you use.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+        call validate_all_in_range_int(kd_indices, n_vectors, ierr, min=1_int32, max=n_vectors)
+        if (is_err(ierr)) return
+
+        call validate_in_range_real(r, ierr, min=0.0_real64)
+        if (is_err(ierr)) return
+
+        M_ALLOCATE(tmp_stack(CM_KD_STACK_ENTRY_SIZE, CM_KD_TRAVERSAL_STACK_DEPTH, n_vectors))
+        M_ALLOCATE(tmp_vicinity_mask(n_vectors, n_vectors))
+
+        call calculate_labels_as_density(vectors, n_dimensions, n_vectors, r, &
```
schroederaaron, line 211: doubled validations, why don't you call the helper directly?

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
...
+        do concurrent(i_vec=1:n_vectors) &
+            shared(vectors, n_dimensions, n_vectors, r, dimension_order, kd_indices, tmp_stack, tmp_vicinity_mask, label_densities)
+
+            ! Call the simplified K-D search wrapper
+            call vicinity_vectors(vectors(:, i_vec), vectors, n_dimensions, n_vectors, r, &
+                                  dimension_order, kd_indices, tmp_stack(:, :, i_vec), &
+                                  tmp_vicinity_mask(:, i_vec))
+
+            ! Accumulate density counts safely using the mask slice
+            label_densities(i_vec) = real(count(tmp_vicinity_mask(:, i_vec)), real64)
```
schroederaaron: Why exactly are we doing this? You effectively run a count function scanning the entire n^2 size array to recover a number you already know. Just return the count from the vicinity_vectors and you save a lot of runtime.

schroederaaron: Matter the fact, you should get rid of the whole `tmp_vicinity_mask` which would not only save you the count operations on the whole n^2 array, but also zeroing the full array at the beginning of the k-d tree function which takes again n^2 as every element is set to .false. , it would also save you the whole memory required for the array, which can grow to gigabytes quickly if you have large inputs.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,305 @@
+#include "macros.h"
+#define CM_KD_STACK_ENTRY_SIZE 3
+#define CM_KD_TRAVERSAL_STACK_DEPTH 64
```
schroederaaron, line 3: Also defined in kd-tree module. Changing one but not the other could cause issues

src/f42_kd_tree.F90
```diff
@@ -188,82 +191,162 @@ end subroutine partial_sort_by_dimension
...
         end if
-        
+
         point_values = points(:, kd_indices(position))
     end subroutine get_kd_point
 
-end module f42_kd_tree
+    !> Finds all reference points within a given radius around a query coordinate vector.
+    !! Returns a logical mask mapping matching reference positions.
+    pure subroutine vicinity_vectors(query_point, points, num_dimensions, num_points, r, &
```
schroederaaron, line 262: Function is added as public, can be called from everywhere by everybody. Please add standalone tests for this function to the mod_test_kd_tree. Also, split into `_alloc` and `_helper` as there is currently no input validation or similar.

schroederaaron (changes requested):

src/f42_kd_tree.F90
Specter1999: We discuss this part, and we decide to make implementation as simple as we can. Since we are in an experimental phase. However, if these parts decide to be in the main part, I will add a necessary layer for it.

src/tox_shatter_cluster_data.F90
Specter1999: @schroederaaron 
Hmm, what would be the solution? Shall I define and move them to macros.h as global macros?

src/tox_shatter_cluster_data.F90
Specter1999: That is correct, and in the current implementation, we have two subroutines. One is for returning vectors, and the other one is just counting.

src/tox_shatter_cluster_data.F90
Specter1999: I forget the reason, but Vivian said I need to validate in the _alloc layer and also in the validation layer.

schroederaaron: I don't think we should do that. I mean it makes no real difference if it is only O(1) checks, but i still think you should call the helper directly. @qwerdenkerXD

src/tox_shatter_cluster_data.F90
schroederaaron: Well you could define them in the header of the kd_tree, which makes them available to every module that uses the kd_tree. I would not put it in the makros file tho. What do you think @qwerdenkerXD

src/f42_kd_tree.F90
schroederaaron: Well it should be split up into alloc and helper before you merge into main.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+
+        ! Calculating each vector distance to mean vector
+        do concurrent(i_vec=1:n_vectors) shared(vectors, tmp_mean_vec, n_dimensions, tmp_distances)
+            call euclidean_distance(tmp_mean_vec, vectors(:, i_vec), n_dimensions, tmp_distances(i_vec))
+        end do
+
+        !Sorting perm according to distance
+        call sort_real_heapsort(tmp_distances, tmp_perm)
+
+        percentile = actual_quant*100.0_real64
```
schroederaaron, line 166: calc_percentile got changed to an input of [0,1] in PR 163, so this function will fail after merge. Please be aware of that, i suggest to change it before you merge into main.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+        end do
+
+        call sort_real_heapsort(density_labels, tmp_perm)
+
+        ! Reverse tmp_perm into sorted_perm
+        do concurrent(i=1:n_vectors) shared(sorted_perm, tmp_perm, n_vectors)
+            sorted_perm(i) = tmp_perm(n_vectors - i + 1_int32)
+        end do
+
+        target_percentile = 100.0_real64 - (actual_t_percentile*100.0_real64)
```
schroederaaron: calc_percentile is updated in PR 163 to input values [0,1], this will fail after merge.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+        real(real64) :: median_ambient, mad_ambient, ensemble_center_density, max_allowed_dev
+
+        call set_ok(ierr)
+
+        ! 1. Calculate ambient median density: median_{i=1..N}(rho_i)
+        do concurrent(i_vec=1:n_vectors) shared(tmp_perm)
+            tmp_perm(i_vec) = i_vec
+        end do
+        call sort_real_heapsort(density_labels, tmp_perm)
+        call calc_percentile(density_labels, tmp_perm, 50.0_real64, median_ambient, ierr)
```
schroederaaron, line 721: update to [0,1] range

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+        else
+            required_cols = actual_t_obs
+        end if
+
+        if (.not. allocated(observables)) then
+            M_ALLOCATE(observables(CM_OBSERVABLE_COUNT, required_cols))
+            observables = 0.0_real64
+        end if
+
+        call compute_ensemble_observable(ensemble_mask, density_labels, n_vectors, &
```
schroederaaron, line 836: directly call the helper here. In `compute_ensemble_observable` you have a `validate_all_in_range` which is a duplicated loop through all `n_vectors` elements

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+        !! Array of starting seed vector indices [[tox_shatter_cluster_data(module):identify_ensemble_seeds_alloc(subroutine)]].
+        real(real64), intent(in), optional :: r
+        !! Search radius threshold for surface growth
+        real(real64), intent(in), optional :: alpha_mad
+        !! MAD multiplier for density compatibility (defaults to 0.5)
+        real(real64), intent(in), optional :: alpha_accept
+        !! Acceptance log2 fold-change threshold (defaults to 0.5)
+        integer(int32), intent(in), optional :: t_observables
+        !! Observable history depth window (defaults to 10)
+        logical, allocatable, intent(out) :: ensemble_matrix(:, :)
```
schroederaaron, line 1113: output arrays should be pre-allocated

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+        !! Number of vectors
+        logical, intent(in) :: ensemble_mask(n_vectors)
+        !! Selection mask for active ensemble members
+        real(real64), intent(in) :: density_labels(n_vectors)
+        !! Density labels for all vectors [[tox_shatter_cluster_data(module):calculate_labels_as_density_alloc(subroutine)]].
+        integer(int32), intent(in) :: candidate_count
+        !! Number of candidates added in current growth step
+        integer(int32), intent(in) :: current_iter
+        !! Current growth iteration index (1-based)
+        real(real64), intent(inout), allocatable :: observables(:, :)
```
schroederaaron, line 805: Is there a specific reason this array is not pre-allocated?

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+
+            call grow_single_seed_helper(vectors, n_dimensions, n_vectors, dimension_order, &
+                                         kd_indices, density_labels, seed_indices(i_seed), &
+                                         r, alpha_mad, alpha_accept, t_observables, &
+                                         tmp_stack(:, :, :, i_seed), tmp_vicinity_mask(:, :, i_seed), &
+                                         tmp_perm(:, i_seed), tmp_abs_diff(:, i_seed), &
+                                         tmp_observables(:, :, i_seed), tmp_current_mask(:, i_seed), &
+                                         tmp_backup_mask(:, i_seed), ensemble_matrix(:, i_seed), local_ierr)
+
+            ierr = max(ierr, local_ierr)
```
schroederaaron, line 1372: if an error occurs you could return earlier here i think

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+        ! Pure do concurrent outer-loop parallelization over independent seeds
+        do concurrent(i_seed=1:n_seeds) &
+            shared(vectors, n_dimensions, n_vectors, dimension_order, kd_indices, &
+                   density_labels, seed_indices, r, alpha_mad, alpha_accept, &
+                   t_observables, tmp_stack, tmp_vicinity_mask, tmp_perm, &
+                   tmp_abs_diff, tmp_observables, tmp_current_mask, tmp_backup_mask, ensemble_matrix) &
+            local(local_ierr) &
+            reduce(max:ierr)
+
+            local_ierr = ERR_OK
```
schroederaaron, line 1362: please use the set_ok function

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1543 @@
...
+
+        integer(int32), intent(in) :: n_vectors
+        !! Total number of ambient vectors
+        integer(int32), intent(in) :: n_seeds
+        !! Total number of grown raw ensemble columns
+        logical, intent(in) :: raw_masks(n_vectors, n_seeds)
+        !! Input matrix of raw boolean ensemble masks [n_vectors x n_seeds]
+        integer(int32), intent(in), optional :: min_intersection
+        !! Minimum points in common to trigger a merge (defaults to 1)
+        logical, allocatable, intent(out) :: merged_matrix(:, :)
```
schroederaaron, line 1495: Is there a reason this is not pre-allocated?

schroederaaron (changes requested):

src/tox_shatter_cluster_data.F90
qwerdenkerXD: The robust approach is to always call the helpers, but keep it as is. The branch is still experimental.

Once the code generator is merged, we will remove the wrappers anyway, as they are generated completely.

src/tox_shatter_cluster_data.F90
qwerdenkerXD: `macros.h` is for general purpose macros, so I fully agree to not putting them into that file.

What about a module-specific macro file? I see two options:
1. name like `<module-name>_macros.h`
2. to keep `macros.h` as the only allowed macro header in src we could create a directory `<module-name>/`, containing:
    - `macros.h`
    - `<module-name>.F90`

src/tox_shatter_cluster_data.F90
Specter1999: In the actual ensemble-growth path the observable array is already preallocated. `obtain_ensembles_alloc` allocates `tmp_observables(5, required_cols, n_seeds)` once together with the other per-seed workspaces, and each seed reuses its 2D slice throughout the growth iterations. The allocatable argument here is only for the standalone `compute_ensemble_observable_alloc` convenience wrapper, so it can be called independently when the caller has not allocated the history matrix yet. It is inout because the observable trajectory has to persist between iterations.

src/tox_shatter_cluster_data.F90
Specter1999: The iterations are designed to be independent and potentially execute concurrently; stopping the whole procedure from one iteration conflicts with that execution model.

So I put it immediately after the concurrent loop.

qwerdenkerXD: the ierr is the error. Why do your helpers return ierr? I checked one and it only sets it ok, nothing else (sry, I'm on my phone and can't check more).

Please check your helpers if they really need the ierr. Dropping them makes them compatible for concurrent loops.

Specter1999: You're right. I checked the helper chain more closely. Some helpers currently take `ierr` even though they cannot actually produce an error and only `call set_ok`, so those `ierr` arguments are unnecessary.

src/tox_shatter_cluster_data.F90
Specter1999: Yes. The exact size of `merged_matrix` is not known before the merge because its second dimension is `n_ensembles`, and `n_ensembles` is determined only after the transitive merging is finished. The _alloc wrapper therefore returns an exactly sized `[n_vectors, n_ensembles]` matrix. For callers that want to preallocate, the regular `merge_ensembles `interface already accepts `merged_masks(n_vectors, n_seeds)` as a preallocated maximum-size output buffer and returns `n_ensembles` telling the caller how many leading columns are valid.

qwerdenkerXD: Allocatable outputs are only fine for Fortran-only procedures. For the C wrappers this is not working, because Fortran owns the memory if it's allocating. For making this work, you would require the worst case (n_vectors, n_seeds) for the matrix and document that n_ensembles is the output size.

Just to keep in mind, it's not urgent. You're still experimenting and don't need C wrappers yet.

Specter1999: @qwerdenkerXD 
Okay, now it makes sense. I was mixing up what is fine for a Fortran-only convenience procedure with what can be exposed through a C wrapper.

I believe the same consideration also applies to this discussion: https://github.com/asishallab-group/Tensor-Omics/pull/143#discussion_r3789193354

Shall I change these two output interfaces now so they are already suitable for future C wrappers, or would you prefer to leave that until we add the wrappers? Apart from these two C-interoperability points, I believe the current implementation is close to being ready for merging, assuming the remaining review points are resolved.

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1637 @@
...
+        call validate_all_in_range_real(vectors, size(vectors, kind=int32), ierr)
+
+        ! Optional parameter range validation
+        call validate_in_range_real(mean_to_other_vecs_dist_quant, ierr, min=0.0_real64, max=1.0_real64)
+        if (is_err(ierr)) return
+
+        M_ALLOCATE(tmp_mean_vec(n_dimensions))
+        M_ALLOCATE(tmp_distances(n_vectors))
+        M_ALLOCATE(tmp_perm(n_vectors))
+
```
schroederaaron: add an ierr check here, M_ALLOCATE can potential return an error, should be checked

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1637 @@
...
+        call validate_all_in_range_real(vectors, size(vectors, kind=int32), ierr)
+
+        ! Array and value range validation checks
+        call validate_all_in_range_int(dimension_order, n_dimensions, ierr, min=1_int32, max=n_dimensions)
+        call validate_all_in_range_int(kd_indices, n_vectors, ierr, min=1_int32, max=n_vectors)
+        call validate_in_range_real(r, ierr, min=0.0_real64)
+        if (is_err(ierr)) return
+
+        M_ALLOCATE(tmp_stack(KD_STACK_ENTRY_SIZE, KD_TRAVERSAL_STACK_DEPTH, n_vectors))
+
```
schroederaaron: M_ALLOCATE could return an error

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1637 @@
...
+        call validate_in_range_int(k_seeding, ierr, min=1_int32, &
+                                   max=n_vectors - 1_int32)
+        if (is_err(ierr)) return
+
+        M_ALLOCATE(tmp_perm(n_vectors))
+        M_ALLOCATE(tmp_distances(n_vectors))
+        M_ALLOCATE(tmp_stack(KD_STACK_ENTRY_SIZE, KD_TRAVERSAL_STACK_DEPTH))
+        M_ALLOCATE(tmp_visited_mask(n_vectors))
+        M_ALLOCATE(tmp_newly_covered_mask(n_vectors))
+
```
schroederaaron: check ierr

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1637 @@
...
+            observables(3, current_iter) = obs_3
+            observables(4, current_iter) = obs_4
+            observables(5, current_iter) = obs_5
+        else
+            observables(:, 1:actual_t_obs - 1_int32) = observables(:, 2:actual_t_obs)
+            observables(1, actual_t_obs) = obs_1
+            observables(2, actual_t_obs) = obs_2
+            observables(3, actual_t_obs) = obs_3
+            observables(4, actual_t_obs) = obs_4
+            observables(5, actual_t_obs) = obs_5
```
schroederaaron: You never test if the array actually has a size of 5 in the first dimension. If i pass in an array with size 4 this will run out of bounds causing a segmetation fault in the worst case

src/tox_shatter_cluster_data.F90
```diff
@@ -0,0 +1,1637 @@
...
+            call calculate_density_radius(vectors, n_dimensions, n_vectors, &
+                                          tmp_mean_vec, tmp_distances, seed_perm, &
+                                          actual_r, ierr=ierr)
+            deallocate (tmp_mean_vec, tmp_distances, seed_perm)
+            if (is_err(ierr)) return
+        end if
+
+        ! Allocate per-seed parallel growth workspaces
+        M_ALLOCATE(tmp_stack(KD_STACK_ENTRY_SIZE, KD_TRAVERSAL_STACK_DEPTH, n_vectors, n_seeds))
+        M_ALLOCATE(tmp_vicinity_mask(n_vectors, n_vectors, n_seeds))
```
schroederaaron: way too large array that is never being fully utilized. If i understand correctly you either read per point or over all columns. So i think you can easily reduce it into a reusable smaller workspace array. Just to clarify, currently, with n_vectors = 5000 and n_seeds = 1350, just this array will try to allocate 100Gb+ memory which will fail with `ERR_ALLOC_FAIL`

Note: I am pretty sure you would need to get rid of the one inner do concurrent to drastically reduce the array. But in my opinion, i would rather have a potentially slower 100Mb work array than a fast array that can't be allocated on most devices.

schroederaaron (changes requested):
In general you sometimes have astronomically large work arrays. They may work with your test, and may work with a small dataset. But please, do a quick calculation and if a single temporary work array requires multiple gigabytes of RAM to be allocated, it is most likely not worth it. I know you often need it for do concurrent. But in my opinion, that tradeoff is not worth it. So, keep an eye out for that. I only flagged one array that has to be changed for sure, but there are more that are really large. Please keep an eye on that.

schroederaaron (changes requested):

qwerdenkerXD (changes requested):

PR #152, opened 2026-07-17
Add JSON serialization, plus current tox flyer version

### LINKED ISSUE #70: Converter: tox_data → Flyer JSON
Write a minimal translator that converts a tox_data ZIP into Flyer JSON. Implementation can be in Lua, Python, R, C, or Fortran (choose what’s fastest to deliver and easy to maintain). 
 
# IMPORTANT
We need a fast, reliable way to translate tox_data content into Flyer JSON. If efficient and accurate conversion is difficult with the current Flyer JSON, we will redefine the Flyer JSON spec to support efficient translation from tox_data.

**Assignees:** @qwerdenkerXD

*Migrated from GitLab: https://gitlab.rlp.net/a.hallab/tensor-omics/-/work_items/69*

qwerdenkerXD:
Closes #70


qwerdenkerXD:
I will resolve conflicts once #135 is merged to avoid duplicate work and follow [the discussed order](https://gitlab.rlp.net/a.hallab/tensor-omics/-/merge_requests/56#note_942303):
>1. merge [PR135](https://github.com/asishallab-group/Tensor-Omics/pull/135) (the updated codebase) into main
>2. rework and merge general documentation branch into main, once we have a clear idea of how to write the manual ([Issue 147](https://github.com/asishallab-group/Tensor-Omics/issues/147))
>3. move this MR to GitHub, so we can also verify the new workflow succeeds
>4. merge main into flyer json, resolve conflicts and add the section for the manual

src/f42/f42_json.F90
```diff
@@ -0,0 +1,554 @@
...
+        integer(int32) :: depth, actual_max_depth
+
+        M_DEFAULT_VAL(max_depth, actual_max_depth, 20_int32)
+
+        depth = 0_int32
+        call serialize_array(json_arr, unit, depth, actual_max_depth)
+    end subroutine serialize_json_array
+
+    !> Serializes tox related data to JSON, compatible with the tox_flyer
+    subroutine serialize_tox_data_as_flyer_json(filename, tissues, n_tissues, family_ids, n_families, centroids, gene_ids, n_genes, genes, gene_to_fam, sorted_gene_to_fam_perm, gene_outliers, gene_species, gene_types, ierr)
```
qwerdenkerXD: **Layering: keep the generic JSON core free of tox/domain deps.**

`f42_json` is a nice generic serializer (`json_value`/`json_array`/`json_object` + `serialize_json_*`), but this module also embeds the tox-specific `serialize_tox_data_as_flyer_json` and its `bind(C)` wrapper, which `use tox_errors`/`tox_conversions`. That makes the generic `src/f42/` layer depend on the `tox` domain layer (an inversion of the usual direction — `tox_*` builds on `f42_*`, not the reverse).

Suggestion: split this file — keep the generic JSON primitives in `f42_json` (in `src/f42/`, no tox deps), and move `serialize_tox_data_as_flyer_json` + `serialize_tox_data_as_flyer_json_c` into a tox module (e.g. `src/tox/...`). That keeps `f42_json` reusable and the dependency graph acyclic.

src/f42/f42_json.F90
```diff
@@ -0,0 +1,554 @@
...
+
+    subroutine serialize_string(string, unit)
+        character(len=*), intent(in) :: string
+        integer(int32), intent(in) :: unit
+
+        integer(int32) :: i_chr
+        character(len=1) :: chr
+
+        write (unit, "('""')", advance="no") 
+        do i_chr = 1, len_trim(string)
```
qwerdenkerXD, line 63: **`len_trim` silently drops trailing whitespace from string values.**

`do i_chr = 1, len_trim(string)` trims trailing spaces off every serialized string (both keys and scalar string values). For the flyer use case this is exactly what you want — it strips the padding of fixed-length Fortran `character` arrays (e.g. `"UNASSIGNED.1  "` → `"UNASSIGNED.1"`). But `serialize_string` is also reached by the *generic* public API (`serialize_json_object`/`serialize_json_array`), where a caller may legitimately want to preserve a trailing space in a string, and here it is silently lost.

If the generic core is split out (see the module-layering comment), consider using `len(string)` in the generic path and doing the `trim` only at the tox/flyer boundary where the fixed-length padding semantics are known. At minimum, document that trailing whitespace is not preserved.

src/f42/f42_json.F90
```diff
@@ -0,0 +1,554 @@
...
+        class(*), intent(in) :: element
+            !! JSON value to serialize, can be either [[json_value(type)]] or any type supported by [[json_value(type)]]
+        integer(int32), intent(in) :: unit
+            !! unit of the file to write to
+        integer(int32), intent(inout) :: depth
+            !! Current depth of traversion
+        integer(int32), intent(in) :: max_depth
+            !! maximum recursion depth for traversion of `element`
+
+        if (depth < max_depth) then
```
qwerdenkerXD, line 191: **`max_depth` exceedance silently emits `null` — consider signalling truncation.**

When `depth >= max_depth`, the value is replaced by `null` with no error/warning. That protects against self-referential cycles (good), but it also means a legitimately deep-but-finite structure gets silently truncated to `null`, producing output that looks valid but is quietly wrong. Since these entry points (`serialize_json_object`/`serialize_json_array`) do not take an `ierr`, the caller has no way to detect it happened.

Suggestion: either surface truncation via an optional `ierr`/logical out-arg, or at least call it out prominently in the doc of the public entry points (the default is 20). Not blocking for the flyer data, which is shallow.

fpm.toml
```diff
@@ -14,7 +14,7 @@ gfortran.build.link = ["loess-netlib", "loess-netlib-drotg", "xxhash", "zip"]
 ifx.build.link = ["loess-netlib", "loess-netlib-drotg", "xxhash", "zip", "iomp5"]
 
 optimization.ifx.flags = "-xHost -align array64byte -qopt-zmm-usage=high -qopt-prefetch=3 -qopt-matmul"
-optimization.nvfortran.flags = "-Mconcur -fopenmp -stdpar=multicore"
+optimization.nvfortran.flags = "-Mconcur -fopenmp -stdpar=multicore -Mbackslash"
```
qwerdenkerXD: **`-Mbackslash` here only covers the `optimization` nvfortran build.**

I added `-Mbackslash` to this line during the main-merge to preserve the intent of the old `build_utils.sh` nvfortran branch (nvfortran treats `\` in a character context as an escape by default, which would corrupt the JSON escaping in `f42_json` — see `serialize_string`). gfortran/ifx are unaffected and need no flag.

The gap: this only applies when the `optimization` feature is active (`TOX_MAX_PERFORMANCE`). Plain and `diagnostics` nvfortran builds still miss it, so JSON output would be wrong there. Since `-Mbackslash` is a correctness (not a performance) flag, it should apply to *every* nvfortran build. Could you place it wherever nvfortran flags are always-on in the new fpm-feature scheme? I left it here rather than guess the schema.

src/f42/f42_json.F90
```diff
@@ -0,0 +1,554 @@
...
+    subroutine serialize_real(real_num, unit)
+        real(real64), intent(in) :: real_num
+            !! real number to serialize
+        integer(int32), intent(in) :: unit
+            !! unit of the file to write to
+
+        if (ieee_is_nan(real_num) .or. .not. ieee_is_finite(real_num)) then
+            write (unit, "('null')", advance="no")
+        else
+            write (unit, "(ES24.16E3)", advance="no") real_num
```
qwerdenkerXD: **Minor: `ES24.16E3` emits a leading blank for non-negative reals.**

The sign position produces a space for `>= 0` values, so numbers come out as `[ 3.0104...E-002, 7.5787...E-002]` (note the leading space after `[` and each `,`). It is valid JSON (whitespace between tokens is allowed) and the tests pin it, so nothing is broken — but it is a bit unusual and inflates the payload. If you want tighter output, format into a temporary and `trim(adjustl(...))` before writing. Purely cosmetic; feel free to ignore.

rcpp/tensoromics_functions.cpp
```diff
@@ -4667,3 +4690,75 @@ NumericVector tox_scaled_distance_quantile_rcpp(
...
+  const int filename_len_c = strlen(filename.get_cstring());
+  const int tissue_len_c = get_max_string_length(tissues);
+  std::vector<char> tissues_c = r_to_c_CharacterVector(tissues, tissue_len_c);
+  const int n_tissues_c = tissues.size();
+  const int family_id_len_c = get_max_string_length(family_ids);
+  std::vector<char> family_ids_c = r_to_c_CharacterVector(family_ids, family_id_len_c);
+  const int n_families_c = family_ids.size();
+  const int gene_id_len_c = get_max_string_length(gene_ids);
+  std::vector<char> gene_ids_c = r_to_c_CharacterVector(gene_ids, gene_id_len_c);
+  const int n_genes_c = gene_ids.size();
```
qwerdenkerXD, line 4725: **No cross-check that the per-gene / per-family arrays have consistent lengths.**

`n_genes_c` is taken from `gene_ids.size()`, `n_families_c` from `family_ids.size()`, etc., and every other array is then forwarded by raw pointer. If a caller passes, say, a `gene_species`/`gene_types`/`gene_to_fam`/`gene_outliers` shorter than `gene_ids` (easy to do from R), the Fortran side indexes up to `n_genes` and reads past the end of the shorter buffer — an out-of-bounds read with no diagnostic. Same concern for `centroids`/`gene_expressions` matrix dimensions vs `n_tissues`/`n_families`/`n_genes`, and for the Python wrapper.

Suggestion: validate the lengths/`nrow`/`ncol` up front in the R (and Python) wrapper and raise a clear R error before calling into Fortran. Cheap, and turns a memory-safety bug into a friendly message.

src/f42/f42_json.F90
qwerdenkerXD: Addressed in c84adf8e: moved `serialize_tox_data_as_flyer_json` and its `bind(C)` wrapper into a new `src/tox/tox_flyer_json.F90`; `f42_json` is now free of tox dependencies. Resolving.

src/f42/f42_json.F90
qwerdenkerXD: Addressed in b914b1ef: with the pointer/bulk-array model, stripping the padding of fixed-length Fortran strings is the intended behaviour, so it is now documented explicitly on `serialize_string` and the public entry points. Resolving.

src/f42/f42_json.F90
qwerdenkerXD: Addressed in 5347095f: added an optional `truncated` intent(out) flag threaded through the serializers, plus a dedicated recursion-limit test. Resolving.

fpm.toml
qwerdenkerXD: Addressed in 210d89cb: `-Mbackslash` now lives on an always-active `nvfortran` fpm feature (not the build script); `build.link` is deferred until nvfortran is verified. Resolving.

src/f42/f42_json.F90
qwerdenkerXD: Addressed in 30769d63 (drop the leading blank) and fed9b0bf (switch to the portable `ES24.16E3` + `trim(adjustl(...))` after ifx and gfortran diverged on the minimal-width `ES0.*`). Output is now byte-identical on both compilers. Resolving.

rcpp/tensoromics_functions.cpp
qwerdenkerXD: Addressed in 5e9d5a01: added up-front length/shape validation in the Python and R wrappers (ValueError / named stopifnot) before the Fortran call. Resolving.

qwerdenkerXD:
Replaced by #222, which migrates the flyer JSON writer onto current main, on the code generator. The generic JSON part is now a streaming writer: no tree, no copies, constant memory.

The flyer app itself (`tox_flyer/`) is not part of #222; only the writer of its JSON format is. The branch `69-flyer-json` is left as it is.

PR #153, opened 2026-07-22
131 add code generator

### LINKED ISSUE #131: Develop script for generating C wrappers and other boilerplate code
# C wrapper generation

To avoid repetetive tasks with creating boilerplate code, we should have a script for generating C wrappers. This ensures consistency across all wrappers.
Therefore, here some keynotes:
 - [x] correct naming for alloc (`foo_c`) and non-alloc (`foo_expert_c`) routines should be applied
- [x] null validation needs to be included for all args
- [x] handling for logical and string needs to be supported
- [x] integer mode arguments need to be received as string inside c wrappers and mapped to respective integer value
    - this needs naming convention for mode 
/method arguments to be recognized as such by the script
    - inside the ford documentation for the argument, there should be also a convention for defining the possible values and their meaning, so the script knows the respective string form
        - if possible, this is not special syntax, maybe just a markdown table with the mapping
    - [x] table syntax with links to module parameters
        - [x] replace values by strings (strip prefix MODE_/METHOD_ and take rest lowercase)
- [ ] optional arguments are not supported for C binding, so for now simply make all required there
    - later we can easily extend the generator logic to accept null pointers as not-passed and update all c wrappers in one go
        - this is actually very important for the inputs of `save_tox_data`, as they are nearly all optional. Currently this function is simply reimplemented on the R/Python side
    - [x] DM_DEFAULT(arg)
        - replace phrase by a recommendation → mandatory argument
    - [x] DM_REQUIRED_IF_PRESENT(arg)
        - allow null
    - [ ] M_DOC_NO_DEFAULT(arg)
        - allow null
        - mainly for zipping in tox archive
- [x] the script should have a modular design to make it easily extensible
- [x] wrappers for functions (not subroutines) should also be supported
- [x] everything passed by reference, no `value` attribute for c wrapper args
- [x] ford documentation should be adopted from wrapped function/subroutine, while the pre-doc comment (`!>`) tells that it's a c wrapper for the respective procedure (links to it)

## R/Python wrapper generation

Once we have all this, we can develop a separate script that reuses some functions or is even included in the above script to generate the Python/R side. Therefore, here some notes:
- [x] we need a naming convention for work arrays, as our Python/R wrappers do the allocation theirselves, so the script needs to recognize all work arrays as such
- [ ] we need some phrase to refer to outputs from mandatory functions like for work array size calculation or retrieving metadata from serialized arrays, so the user knows where to get such things from and the Python/R function can call the respective functions to simplify the user's pipeline. The LaTeX documentation for Fortran will include this as well. And if the parsing step recognizes which functions rely on a certain function X, this information can be included as well, like `Needed for functions...`
- [x] Python/R documentation gets their documentation from ford comments (syntactically translated)
    - [ ] Ford-specific syntax needs to be translated (links etc.)
    - [x] macros in comments need to be resolved → operate on preprocessed files
- [x] dimension arguments are inherited from inputs that have the respective dimension size
- [x] shape arguments are inherited from input they are related to
- [x] `n_selected_<arg>` args for `<arg>_selection_mask` args are inherited from the mask
- [x] the usual R/Python wrapper conventions need to be satisfied
- [x] some syntax/phrase in Ford is needed for output sizes, as we have in paralog analysis (n_results) or compute_edf (n_unique) when the actual output size is result-dependent. Having this allows to return arrays sliced.
- [x] some syntax/phrase in Ford is needed for default values of optionals, but here we need two variants
    - [x] default value
        - specify the default for some thresholds etc or random seed, such scalars
    - [x] required if
        - sometimes we have different modes with different requirements of work arrays and some extra inputs (e.g. in paralog analysis the subfunc args or in gene centroids the ortholog_set). In this case we need to tell that in some consistent way
        - *BUT*, also possible: Don't allow conditional arguments for non-helpers, so each case gets its own function that calls the helper correctly (done this way in paralog analysis). This might be a better API
 - [x] such syntax/phases could be preprocessor macros, so they can be verbose without bloating the code

### Rcpp
 - [ ] respect intent attributes
    - external interfaces use `const arg*` for `intent(in)`, else just pointer `arg*`
    - the function declarations should receive values:
        - by immutable reference `const arg&` for `intent(in)`
        - by reference `arg&`  for `intent(inout)` (Usually not the case for us, but exists)

This will ensure a consistent API with consistent error handling etc. and can be extended easily.

**BUT** test functions still need to be created manually (easy LLM task, just give the Fortran tests and the R/Python functions as context and let it generate it, debug a little bit and be happy).

### $\LaTeX$ documentation generation

Having the above generators, there is not much missing to generate the general documentation from our code. The scripts parsed all input arguments, functions and Ford comments, so actually everything is set up to generate the general documentation as well, it seems, but here some notes:
- we need a clear specification on what we need in the documentation. It hasn't been made very clear in the past, only phrases like `it explains general usage... has some examples... gives more information about usage`, but nothing really concrete
    - in my opinion we would really benefit from a very clear specification here that we apply to our Ford documentation then (Ford supports markdown, so showing embedded examples etc. is everything possible)
    - this would result in consistent and detailed documentation directly inside the codebase and thus in the generated Ford html pages as well, while our custom generator script can parse the Ford comments and generate our $\LaTeX$ pages consistently (which is also insane, as we only need to update the script in one place if we want to improve the pages or just make them prettier)
- as touched in the above note, we need a convention for Ford documentation once having a clear idea of what we need in the documentation (will become clear when developing this, but a consistent structure like description>examples may make sense)
- Ford-specific syntax needs to be translated
    - links can be hyperref links to the tagged section of the linked element
- macros in comments need to be resolved → operate on preprocessed files

# Considerations

- I see two ways for generating the code:
    1. specify the file with the subroutine definitions to generate the code for and specify the concrete subroutines/functions
    2. (recommended) Every subroutine sets the `category` meta-tag from Ford like `caregory: C-interface`, so the script automatically detects procedures to generate code for
        - advantage here is that all our wrappers are always consistent if we change some things in our logic
        - Ford currently doesn't show the categories in the generated doc, but it's planned to show sth like a list in future, so one could inspect the C-interface category and sees all subroutines having a c wrapper. Not too bad, I think
        - disadvantage, but worth straight-forward one-time workload, all current procedures with c wrapper need this tag (actually no problem)
- the generator scripts should always throw an error if something is incorrectly defined (e.g. missing mapping for mode arguments or missing example block in Ford, needed for documentation generation) (***OR***, also develop an extra validation script that goes through the Project object and validates as mich as possible, but nonetheless throwing errors in generation is also helpful)
    - we can add the script to the workflow, which will fail for such inconsistencies
    - still, incorrectly named work arrays cannot be detected as incorrect, as it might be a non-temporary real output, so just throw errors if applicable, so when incorrect definitions can be implied etc.

The generator should be developed in multiple parts:
1. parser -> parses all structures in our codebase (function/variable declarations, Ford comments) into Objects of classes like `Variable`, `Procedure`.
    - used from Ford API
    - maybe add a valudation script that validates things, like:
        - missing Ford comments
        - missing defaults for optionals
        - naming conventions like dimension arg names
        - missing ierr in non-helpers (except utils)
        - inconsistencies in array declaration (should use dimension attribute, should use explicit size in non-helpers (except utils))
2. C wrapper generator
    - we could have one c wrapper dir `src/c_interface` with `<fortran_modname>_c.F90` for all our modules (could also mirror our file hierarchy, but not necessarily)
    - Each module contains all its c wrapper routines, so not extern as now, but contained by a module (works because the bind(C, name="...") is the actual exposing magic). This way we have them grouped in Ford as well and a clear interface (module contains Fortran, module_c the c wrappers)
    - the modules could be wrapped into an `#ifndef NO_C_INTERFACE` condition, if someone doesn't need the wrappers
    - anyway, this way the Fortran code files don't need to be modified and we have clear c wrapper modules
3. Language interface: each interfacing language get an own module, implementing a generator function that takes `snippets_output_dir`, `wrapper_output_dir` and `c_wrappers`
    - The snippet serialization is one helper function that expects the body, the docstring and a language specific prefix
4. LaTeX generator
    - the main work here will be to translate Ford syntax (Markdown, links, `@note`) into LaTeX documentation equivalents
        - check how Markdown is being processed by Ford (maybe there is some parser as well, but might be that it directly produces HTML)
            - would be helpful for e.g. table parsing

### LINKED ISSUE #170: Check option to use C bool and C char instead of Fortran kinds to avoid any copy between C <-> Fortran
Check whether all functions that we can use on Fortran native logical and Fortran native char would still work, if we used the C alternatives.

qwerdenkerXD:
This PR adds the code generator to our repo, please leave yout thoughts.

Closes #131
Requires #135 to be merged

---

## What changed under `src/`

The generator is the tooling; this is what it did to the sources. Everything below is
hand-written unless it lives under `src/generated/`.

### The `_impl` layer

Every module that exports to C is now split in two. The hand-written half keeps the
algorithm and is named `<name>_impl.F90`; the public half is generated from it into
`src/generated/`. The old hand-written triple `foo_helper` / `foo` / `foo_alloc` is gone,
and the suffixes were renamed to say what the procedure is rather than what it does
internally:

| before | now | who writes it |
|---|---|---|
| `foo_helper` | `foo_impl` | author |
| `foo_alloc` | `foo` — the recommended API, allocates its own workspace | generator |
| `foo` | `foo_expert` — caller supplies the workspace | generator |

25 `_impl` modules produce 21 public Fortran modules with 101 public procedures, 30 of
them `_expert`. A generated module publishes the `_impl` module's own documentation, so
the doc comment an author writes is what Python, R and the C header show.

### Input validation is generated, not hand-written

Valid ranges are now documented on the argument (`DM_MIN`, `DM_MAX`, `DM_SENTINEL`,
`DM_ALLOW_NAN`, `DM_ALLOW_INFINITE`) and the wrapper emits the check — about 500
`validate_*` calls in total, with the right `arg_pos` for each emitted signature.
Finiteness is the default contract for every real argument; an argument that may be NaN
or infinite has to say so. `tox_errors` grew `allow_nan` / `allow_infinite` on the two
real validators to support that.

### `logical(c_bool)` everywhere

152 declarations across 51 files. Any logical that crosses the C boundary — or might
later — is `logical(c_bool)`, so no wrapper has to marshal one kind into the other.
Benchmarked first (`misc/bench/logical-kinds/`): the copy we removed is small but real
(13 µs for a 20 000-element mask), `c_bool` inside a kernel is *cheaper* than default
`logical` at large n because it moves a quarter of the bytes, and the marshalling buffer
was an automatic array large enough to blow ifx's stack at 10M elements. `real64` and
plain `character(len=*)` stay as they are; the safeguard is what licenses that.

### `f42_safeguard`

`src/safeguard.F90` → `src/f42/f42_safeguard.F90`, and from 3 compile-time guards to 8:
four kind *equalities* (`c_int`/`int32`, `c_double`/`real64`, `c_double_complex`/`real64`,
`c_char`/default) and four *existence* checks (`c_bool`, `c_size_t`, `c_int64_t`,
`c_signed_char`), since `iso_c_binding` reports an absent kind as `-1` rather than
failing.

The module depends on nothing but the intrinsics, deliberately. A dependency would lower
its compile priority and some other module would fail first, with the compiler's message
instead of ours. **Any module naming a C kind must `use f42_safeguard`** — that is the
only way to order a Fortran build. 71 modules do, and the generator now checks it. Eight
`TEST_KIND_MISMATCH_*` directives prove each guard actually fires.

### Layout

| move | reason |
|---|---|
| `src/config.F90` → `src/f42/f42_config.F90` | it is f42 infrastructure |
| `src/tox/tox_errors.F90`, `tox_conversions.F90` → `src/f42/` | used by f42 itself, so they belong below it |
| `src/f42/f42_utils.F90` → six modules under `src/f42/utils/` | math, random, sort, stats, vector, plus a `f42_utils_impl` umbrella so one `use` still reaches everything |
| `src/tox/data/integration/**` → `src/tox/data_integration/*_impl.F90` | the parent-interface + submodule structure is dissolved; the wrappers regenerate |
| new `src/generated/` | `tox/`, `f42/`, `bindings/c/` (36 modules), `bindings/r/` (36 pure-C `.Call` shims + `init.c` + `tox_marshal.h`, replacing Rcpp — see `misc/bench/r-binding-backend/`) |

`src/generated/**` is marked `linguist-generated` in `.gitattributes`, so a review sees the
`_impl` change rather than the fan-out it produced. `.gitignore` was reworked to anchor
every whitelist rule at the repository root.

### `src/macros.h`

12 macros → 28. The additions are the annotation contract the generator reads
(`DM_MIN`, `DM_MAX`, `DM_SENTINEL`, `DM_ALLOW_NAN`, `DM_ALLOW_INFINITE`, `DM_DEFAULT`,
`DM_OUTPUT_FROM`, `DM_REQUIRED_IF_MODE`, `DM_RESULT_SIZE_IS`, `DM_PROLOGUE`) plus
`M_EXPORT_C` (the export marker — a procedure without it is not bound),
`M_IMPLICIT_NONE` (`implicit none (type, external)`), `M_CHECK_ARRAY_NON_NULL` and
`M_ERR_ARG_POS_FACTOR`. Nothing was removed.

### What the generator refuses to build

Running the generator is also a check on the sources — close to 70 of them, six of which
are warnings and the rest fatal. They are collected and reported together, as diagnostics
about the Fortran rather than a Python traceback:

```
error: module 'f42_serde_arrays_serialize_logical' names 'c_bool' without using 'f42_safeguard'
  --> src/f42/serde/arrays/serialize/f42_serde_arrays_serialize_logical.F90:4
  module 'f42_serde_arrays_serialize_logical' in project 'TensorOmics'
  note: add 'use f42_safeguard'. Its guards report which C kind a platform is missing, and
  they only get to speak if it is compiled first -- which a dependency is the only way to
  arrange. [...]
```

Roughly grouped:

- **C-kind ordering** — a module naming a C kind must `use f42_safeguard`.
- **Expressible in C** — explicit `intent`; no derived types, `allocatable`/`pointer`
  dummies or deferred-length `character(len=:)`; a rank limit on character arrays; kinds
  with a known C mapping; an `ierr` that is a scalar `integer(int32), intent(out)`.
- **Documentation is API** — every `[[...]]` Ford link must resolve to something that
  exists and is published. A dangling link is a build error, not just a broken doc page.
  Missing summaries and undocumented arguments are the warnings.
- **`_impl` conventions** — an implementation module sits at the mirrored path, is not
  exported itself, does not allocate (that is the wrapper's job), does not `use` a
  generated module, and is not named for the wrapper it generates.
- **Annotations have to hold up** — a `mode` argument needs exactly one table of accepted
  values and `DM_DEFAULT` must name one of them; `DM_OUTPUT_FROM` must name a real
  exported producer whose inputs match by name; `DM_RESULT_SIZE_IS` must point at a scalar
  integer; `<arg>_shape`, extent and `tmp_` arguments must have the type and intent their
  name promises.
- **`DM_PROLOGUE` ordering** — the prologue exists and its signature matches, and it may
  not write a value the wrapper already used to size, sort or allocate something.

`python helper/generate_code.py --check` runs all of it and additionally fails if
`src/generated/` is stale; it is a CI step, so neither a broken annotation nor a
forgotten regeneration can reach `main`.

---

## Diagnostics, and what came out of turning them on

CI now runs both compilers with `--diagnostics` as well as `--max-performance`, and the
gfortran leg had been compiling 366 files while reporting nothing — `-fcheck=all` is
runtime-only, so it had no warning flags at all. It gained
`-Wall -Wextra -Warray-temporaries -Wrealloc-lhs-all -std=f2023 -pedantic`; ifx gained
`-stand f23`. `-std=f2023`, not `f2018`, because the code uses `do concurrent … reduce(+:)`.

That surfaced two things worth naming:

**82,000 array temporaries in the tests.** Every assertion helper takes explicit-shape
dummies, so a vector subscript, an elementwise expression or a matrix *row* has to be
gathered into a temporary first. 81,960 of them came from one line in `mod_test_tox_data`.
A `severe (408)` hid the rest: `mod_test_quantile_normalization`'s insertion sort wrote
`j >= 1 .and. a(j) > key`, and Fortran does not short-circuit `.and.` — the run died at
suite 9 of 33, so `tox_data` was never reached.

**Declaration order.** `real(real64) :: array(n)` above `integer(int32) :: n` is a GNU
extension that `-std=` rejects; five hand-written files had drifted from a rule both Fortran
emitters had followed since they were written. The generator now enforces it on the sources
too (`check_extents_are_declared_first`), for extents and character lengths, over every
procedure rather than only exported ones — this is about whether the file compiles.


python/test_helpers.py
```diff
@@ -79,5 +80,5 @@ def assert_error(func, msg):
         assert False, msg
     except AssertionError as e:
         raise e
-    except (RuntimeError, ValueError, FileNotFoundError) as e:
+    except (RuntimeError, ValueError, TypeError, FileNotFoundError) as e:
         pass
```
qwerdenkerXD: >Just a general thought, not necessarily relevant for this PR. Wouldn't it make sense to also let the assert_error() function accept a certain error code, e.g.
>
>assert_error(lambda: function(something), 201, "Test failed, expected ....") or even with textual form of the error code? That way we wouldn't only assert that some error is raised, but that the correct error is raised, which may be relevant for subroutines with multiple possible error codes

qwerdenkerXD:
@schroederaaron @Specter1999 @asishallab The code generator is ready for review and merging.

The proposed C layer (have functions in C that allocate outputs and are the ones called from Python/R) is not implemented.
It is a good follow-up candidate, as it changes only the code emission, not the parsing.

qwerdenkerXD:
### The Python emitter drops a procedure's doc body; roxygen keeps it

Python callers get strictly less documentation than R callers from the same Fortran source, for every generated procedure.

`render_roxygen` writes the summary and then the body:

```python
title = _render(procedure.meta.summary or wrapper.stripped_name, "r")
writer.line(f"#' {title}")
writer.line("#'")

resolver = getattr(emitter, "links", None)
notes = _notes(wrapper.doc, resolver)
if notes:
    for line in notes:
        writer.line(f"#' {line}" if line else "#'")
```

`render_numpydoc` writes the summary and goes straight to `Parameters`:

```python
writer.line('r"""' + _render(wrapper.procedure.meta.summary or wrapper.stripped_name, "python"))

inputs = emitter._inputs(wrapper)
outputs = emitter._outputs(wrapper)

if inputs:
    writer.blank()
    writer.line("Parameters")
```

Everything an author writes in the `!|` block after `summary:` reaches R and C, and is dropped for Python.

### Why it is worth fixing rather than living with

The body is usually where the procedure's contract actually is. `compute_tissue_versatility` is a clean example — the direction of the scale lives only there:

R:
```r
#' Computes normalized tissue versatility for selected expression vectors.
#'
#' The metric is based on the angle between each gene expression vector and the space diagonal.
#' Versatility is normalized to [0, 1], where 0 means uniform expression and 1 means expression in only one axis.
```

Python:
```python
r"""Computes normalized tissue versatility for selected expression vectors.

Parameters
----------
```

A Python caller is never told whether 1 means uniform or means single-tissue — the one thing they need in order to read the number. Same shape of loss on `normalization_pipeline` (which steps it performs, in which order) and on `calc_fchange` (that its input must already be on a log2 scale, or the subtraction is not a fold change).

Numpydoc has the slot for it: free text between the summary line and `Parameters` is the extended summary, so this is a placement question rather than a format constraint.

Found while writing the manual's recipes; not urgent, and deliberately not fixed here since it would touch every generated Python file.

helper/codegen/emit/c_call.py
```diff
@@ -0,0 +1,878 @@
...
+from __future__ import annotations
+
+import re
+
+from ..abi.model import CArgument, Conversion, CWrapper, CWrapperModule, Origin
+from ..ir.types import BaseType, Intent
+from ..render import Writer
+
+#: The R shared object this binding is registered into (drives `R_init_<name>`).
+R_DLL_NAME = "tensoromics"
```
schroederaaron: i think this will never match anything as our shared object is called `libtensor-omics.so`, right? So `R_INIT_TENSOROMICS` never runs i think.

helper/codegen/emit/c_call.py
```diff
@@ -0,0 +1,878 @@
...
+    memset(buf, ' ', total);
+    return buf;
+}
+
+// char[len * n] fixed-width -> STRSXP, each slot with its trailing blanks removed: Fortran
+// blank-pads whatever it assigns into a character(len=n), and the wrapper hands that buffer
+// straight through. Trailing NULs are deliberately not stripped -- nothing writes them any
+// more, and Rf_mkCharLen turning a stray one into a loud R error is the right answer.
+// Returned unprotected: the caller protects it straight into a result slot.
+static inline SEXP tox_char_out(const char* buf, int len, int n) {
```
schroederaaron: tox_char_out fills a freshly allocated STRSXP with Rf_mkCharLen while it is unprotected; a GC mid-loop can collect it. Reached by read_tox_data_into() with many gene ids. Generated copy: [tox_marshal.h:111](vscode-webview://1gej8430k878a7q00i50lgqmgbsulfc9o139jbusrjtj2ibvlplo/src/generated/bindings/r/tox_marshal.h#L111).

helper/codegen/emit/error_arguments.py
```diff
@@ -0,0 +1,37 @@
...
+        # call, so by the time Fortran can complain they agree and naming the first is true
+        return roles.extent_of[0].name
+    if roles.shape_of is not None:
+        return roles.shape_of.name
+    if roles.mask_count_of is not None:
+        return roles.mask_count_of.name
+    return None
+
+
+def sources_of(procedure: Procedure) -> list[str | None]:
```
schroederaaron: source_of blames an output for an error on a caller-supplied extent (an R error about n_families reports argument 'dscale').

src/generated/bindings/c/tox_data_archive_c.F90
```diff
@@ -0,0 +1,477 @@
...
+        M_CHECK_NON_NULL(zip_filename_strlen)
+        M_CHECK_NON_NULL(keys_strlen)
+        M_CHECK_NON_NULL(n_keys_elements)
+        M_CHECK_NON_NULL(filenames_strlen)
+        M_CHECK_NON_NULL(n_filenames_elements)
+        M_CHECK_ARRAY_NON_NULL(zip_filename, zip_filename_strlen)
+        M_CHECK_ARRAY_NON_NULL(keys, keys_strlen * n_keys_elements)
+        M_CHECK_ARRAY_NON_NULL(filenames, filenames_strlen * n_filenames_elements)
+
+        call c_f_pointer(c_loc(zip_filename), zip_filename_f)
```
schroederaaron: c_loc of a zero-sized character buffer when strlen == 0 (R passing "") — precisely the case M_CHECK_ARRAY_NON_NULL exists to prevent.

schroederaaron (changes requested):
Besides from what claude suggests, i would probably rework the codegen_guide.md a bit. It currently holds a lot of section explaining stuff with actual numbers. e.g. something like 37 out of 82 subroutines use this. It is interesting to see those numbers, but every single addition/removal of a function makes those statements invalid. So, i would prefer to rework them either to like, "around 45%", or remove them in total, as they are not needed for users anyways. Also, please remove the "§" smybols from the guide. It is a good way to refer to sections, but i am not sure if every device can display it properly. Just replace every "§" with "Section" and every "§§" with "Sections" and we should be good. 

Generated code looks good to me, all edge cases i could think of are covered i think.

helper/codegen/emit/error_arguments.py
qwerdenkerXD: this is because n_families is not being passed by the caller. Empty input will always blame the array, not its size argument a user never passes explicitly

src/generated/bindings/c/tox_data_archive_c.F90
qwerdenkerXD: I don't get this. What does a null check prevent for empty strings?

schroederaaron: Right — the null check prevents nothing here, and it isn't supposed to. That part is fine.

The problem is the other c_loc. Two lines are involved in tox_data_archive_c.F90:


M_CHECK_ARRAY_NON_NULL(zip_filename, zip_filename_strlen)   ! 71 - guarded by if (N > 0), correctly skipped
call c_f_pointer(c_loc(zip_filename), zip_filename_f)       ! 75 - not guarded, runs anyway

Line 71 does exactly what your comment describes. Line 75 then performs the very c_loc on a zero-size array that the guard exists to avoid — and now with an unchecked pointer. F2018 18.2.3.6 requires an array argument to C_LOC to have nonzero size, so this is non-conforming.

It's reachable from R with no exotic input: tox_max_strlen("") is 0, tox_char_in(x, 0) calls R_alloc(0, 1) which returns NULL, and .tox_as_character doesn't check length. I built a copy of the emitted pattern and called it from R:


R shim: strlen = 0, buffer pointer = (nil)
Fortran runtime error: Pointer actual argument 'zip_filename_f' is not associated
exit=2
That's with -fcheck=all, i.e. our own diagnostics.gfortran.flags in fpm.toml:21 — the R session dies and ierr never comes back. Without checks it silently returns ierr = 0 and a len-0 string. And with a valid non-null pointer but zero extent, gfortran handed back a stack address instead of the caller's buffer, so it isn't only a conformance nit.

Note the Python emitter already floors this: .ljust(1) for scalars, max(...) or 1 for arrays (tox_data_archive.py:162-166). The R side has no equivalent.

Two fixes, and I think we want both:

Floor tox_max_strlen at 1, mirroring Python. Closes the "" route cheaply.
Guard the remap in _remap_local (fortran_c.py:391) the same way the check is guarded, so the emitted pattern is conforming for direct C callers too. Not a validation that rejects zero length — that would contradict the "callee's validate_dimension_size decides" policy and would break legitimate character(0) vectors, where strlen * n is 0 regardless of width. Just skip the c_f_pointer and leave the local pointing at a zero-length target.
The floor alone doesn't cover character(0), so they're complementary rather than alternatives.

schroederaaron (changes requested):

helper/codegen/emit/c_call.py
qwerdenkerXD: Right, it never ran. R enters a library through `R_init_<basename>` — `lib` prefix included, no substitution at all — so `libtensor-omics.so` needed `R_init_libtensor-omics`, which is not a C identifier. On the shipped library: 0 registered routines, `dynamicLookup` still TRUE. The wrappers only worked because `.Call("<name>_call")` falls back to dynamic lookup.

A quoted `__asm__("\"R_init_libtensor-omics\"")` label does work under gcc, but clang and icx put the quote characters *into* the symbol name, silently — and fpm compiles the shims with icx whenever the Fortran compiler is ifx, so half of CI would have gone on not registering.

Fixed in 2e58534a by renaming the fpm package to `tensor_omics`: `libtensor_omics.so`, entered through `R_init_libtensor_omics`. Now 108 routines registered and `dynamicLookup` FALSE, on gfortran and ifx builds. Before switching it on, all 108 `CallEntries` arities were checked against the C definitions and the R call sites (0 mismatches). `test_end_to_end_r` asserts the registration, and that `R_DLL_NAME == "lib" + <fpm name>`.

helper/codegen/emit/c_call.py
qwerdenkerXD: Agreed — fixed in 60e59b79: `out` is protected across the fill and unprotected before returning, so the caller's `PROTECT` is still the one that keeps it.

Honest caveat: I could not make it fail at runtime. `gctorture(TRUE)` does collect the vector, but the freed node is only reused by an allocation of its own size class, and the loop allocates CHARSXPs — a 400-string round trip came back intact with and without the fix. So it's a violation of R's protect-before-allocating rule rather than an observed crash, and the guard is a unit test on the emitted header (mutation-checked). `tox_bool_out` and `tox_shape_of` stay unprotected, since their single allocation is their last one; a test pins that property too.

src/generated/bindings/c/tox_data_archive_c.F90
qwerdenkerXD: You were right, and thanks for the repro — I reproduced it through the built library as well. It's worse on ifx: gfortran `-fcheck=all` gives the "Pointer actual argument is not associated" error, ifx `-check all` segfaults. The non-NULL, zero-extent case is harmless on both (`len=0`, no trap), so the live failure is specifically the NULL that `R_alloc(0, 1)` returns.

Both of your fixes, in 38b5a8ee, with one correction to the first: the floor sits inside the *present* branch of `tox_max_strlen` only. `R_NilValue` has to keep reporting 0 — width 1 beside a null pointer would make an omitted optional come back `ERR_EMPTY_INPUT`.

The remap guard is a new `M_CHECK_CHARACTER_VIEW`, and it refuses an empty buffer with `ERR_EMPTY_INPUT` instead of deferring. That is the one exception to `M_CHECK_ARRAY_NON_NULL`'s "let the callee decide": a string is a view built before the call, so there is no address to view and no way to defer — the verdict is the one `validate_dimension_size` would give for a zero extent. After the fix: `""` → a one-blank string (the callee says "could not open file"), `character(0)` → "empty input arrays", `c("", "")` → succeeds. None fatal.

PR #161, opened 2026-07-30
Add angle-based outlier detection

### LINKED ISSUE #88: Angle-Based Outlier Detection for Directional Shifts
***In GitLab by @asishallab on Oct 27, 2025, 17:12 GMT+1:***

# Angle-Based Outlier Detection for Directional Shifts

## Goal

Detect duplicated genes whose expression has diverged in **direction** rather than **magnitude**.

In Tensor Omics, **shift vectors** Δᵢ = pᵢ − o describe how each derived expression vector (*pᵢ*) differs from its ancestral state (*o*, ortholog or family centroid).

So far, outlier detection has focused on the **lengths** of Δ (RDI → magnitude shifts).

Here, we introduce an **angle-based method** to identify directional outliers — cases where a duplicate pivots into a different expression domain, even if its total magnitude remains similar.

---

## Core Angle Metrics

For each family **F**, copies *pᵢ*, and ancestor *o*:

### (a) Angle to ancestor (ortholog)
$$
\theta_i^{\text{orth}} = \arccos \frac{p_i \cdot o}{\|p_i\| \, \|o\|}
$$
Measures how far the derived copy has rotated relative to its ancestral expression.

### (b) Angle between source and copy (optional)
$$
\theta^{\text{sc}} = \arccos \frac{s \cdot c}{\|s\| \, \|c\|}
$$
For cases where a specific **source** paralog and its **copy** are defined.

### Relative Axis Plane

Implement the method and measures also for angles measured in the RAP.

---

## Why Circular Statistics

Angles live on a **circle**, not a line.  
Using standard linear SDs would break near the boundaries $0$ and $\pi$.  
We therefore use **circular variance** and **circular standard deviation**, which correctly respect the geometry of angular data.

For unit shift directions $u_i = \Delta_i / \|\Delta_i\|$:

$$
R_F = \Big\|\frac{1}{m}\sum_{i=1}^m u_i\Big\|, \quad
V_F = 1 - R_F, \quad
s_{c,F} = \sqrt{-2\ln R_F}.
$$

- **$R_F$** — mean resultant length (measure of concentration).
- **$V_F$** — circular variance.
- **$s_{c,F}$** — circular standard deviation of shift angles within the family.

Tightly conserved families have small $s_{c,F}$, while divergent families have larger $s_{c,F}$.

---

## Normalization of Angles

To make angle outlier detection **comparable across families** with different inherent dispersions, we **scale each copy’s angle** by its family’s circular SD:

$$
z_i = \frac{\theta_i^{\text{orth}}}{s_{c,F}}
$$

This is analogous to **voom-like variance normalization**:
- families with naturally large angular spread are down-weighted,
- conserved families highlight meaningful outliers.

This **stabilizes the null distribution of angles** across gene families.

---

## Outlier Detection

We pool the scaled angle scores $z_i$ across all families and build an **empirical cumulative distribution function (ECDF)**.

- Quantile estimation: **Hyndman–Fan type 7** linear interpolation (R/NumPy default).
- Threshold: upper percentile (e.g. 95%) of the ECDF.

$$
\text{outlier if } z_i \ge q_{0.95}
$$

Optional: control FDR when many copies are tested.

---

## Biological Interpretation

For detected angle outliers:

- Project ancestor and duplicate to the **RAP plane** (orthogonal to the space diagonal).
- Inspect axis-wise differences
  $$
  \delta^{\text{RAP}} = p_i^{\text{RAP}} - o^{\text{RAP}}
  $$
  to identify **which tissues or domains were gained or lost**.
- This yields an interpretable “expression pivot” signature.

---

## Tasks

- [ ] Implement angle calculation (a) and (b) in kernel  
- [ ] Add circular variance & SD calculation to analysis core  
- [ ] Normalize angles by family circular SD  
- [ ] Implement ECDF with Type 7 interpolation for thresholding  
- [ ] Integrate outlier call into tensoromics pipeline  
- [ ] Add method description to the LaTeX methods section

---
R code already implemented, you can just convert this to fortran. Please make sure with our existing subroutines that we are not duplicating any function:

```
tox_angle_to_centroid <- function(genes, centroids, gene_to_family) {
  n_genes <- ncol(genes)
  d <- nrow(genes)

  # Extract corresponding centroid for each gene
  matched_centroids <- centroids[, gene_to_family, drop = FALSE]

  # Compute dot product between columns
  dot_prod <- colSums(genes * matched_centroids)

  # Calculate vector norms
  norm_genes <- sqrt(colSums(genes^2))
  norm_centroids <- sqrt(colSums(matched_centroids^2))

  cos_theta <- dot_prod / (norm_genes * norm_centroids)
  cos_theta <- pmin(pmax(cos_theta, -1), 1)  # Numerical clamping to avoid acos domain errors
  angles <- acos(cos_theta)

  return(angles)
}


tox_circular_sd_per_family <- function(genes, centroids, gene_to_family) {
  d <- nrow(genes)
  n_genes <- ncol(genes)
  n_families <- ncol(centroids)

  # Calculate difference vectors Δ_i = p_i - c_fam
  delta <- genes - centroids[, gene_to_family, drop = FALSE]

  # Normalize Δ_i to unit length (unit vectors)
  norms <- sqrt(colSums(delta^2))
  # Avoid division by zero
  norms[norms == 0] <- NA_real_
  unit_vectors <- sweep(delta, 2, norms, "/")

  # Initialize results
  R <- numeric(n_families)
  s_c <- numeric(n_families)

  for (f in seq_len(n_families)) {
    idx <- which(gene_to_family == f)
    if (length(idx) <= 1) {
      R[f] <- NA_real_
      s_c[f] <- NA_real_
      next
    }

    # Mean of unit vectors
    mean_vec <- rowMeans(unit_vectors[, idx, drop = FALSE], na.rm = TRUE)
    R[f] <- sqrt(sum(mean_vec^2))

    # Circular standard deviation
    if (R[f] > 0) {
      s_c[f] <- sqrt(-2 * log(R[f]))
    } else {
      s_c[f] <- NA_real_
    }
  }

  return(list(R = R, s_c = s_c))
}

# angles: vector of length n_genes (radians 0–pi), may contain NA
# s_c:    vector of length n_families (circular SD per family)
# g2f:    gene_to_family (length n_genes), indices 1..n_families
tox_scaled_angles <- function(angles, s_c, g2f, eps = 1e-12) {
  stopifnot(length(angles) == length(g2f))
  # Extract the s_c value for each gene's family
  sc_per_gene <- s_c[g2f]
  # Avoid division by zero (families with R≈1 → s_c≈0)
  sc_per_gene[!is.na(sc_per_gene) & sc_per_gene < eps] <- NA_real_
  # z_i = theta_i / s_c_F
  z <- angles / sc_per_gene
  # If angle is NA or s_c is NA → z will be NA (correct behavior)
  z
}

# z: vector of angular z-scores (may contain NA)
# p: percentile in [0,1], e.g. 0.95 for 95th percentile
tox_angle_outliers <- function(z, p = 0.95) {
  z_valid <- z[is.finite(z)]  # Remove NA/Inf if present
  if (length(z_valid) == 0L) {
    return(list(
      threshold = NA_real_,
      is_outlier = rep(FALSE, length(z)),
      z = z
    ))
  }
  thr <- as.numeric(stats::quantile(z_valid, probs = p, na.rm = TRUE, type = 7))
  is_out <- (z >= thr) & is.finite(z)
  list(threshold = thr, is_outlier = is_out, z = z)
}


#' Calculate angle (in radians) between two vectors
#'
#' Computes the angle between two numeric vectors of the same dimension.
#' Returns a value in [0, pi] radians.
#'
#' @param vec1 First numeric vector
#' @param vec2 Second numeric vector (same length as vec1)
#' @return Numeric value: angle in radians between vec1 and vec2
#' @examples
#' angle <- tox_angle_between_vectors(c(1,0), c(0,1)) # pi/2
#' angle <- tox_angle_between_vectors(c(1,1), c(1,-1)) # pi/2
#' angle <- tox_angle_between_vectors(c(1,0,0), c(0,1,0)) # pi/2
#' 
tox_angle_between_vectors <- function(vec1, vec2) {

  dot_prod <- sum(vec1 * vec2)
  norm1 <- sqrt(sum(vec1^2))
  norm2 <- sqrt(sum(vec2^2))
  if (norm1 == 0 || norm2 == 0) {
    return(NA_real_)
  }
  cos_theta <- dot_prod / (norm1 * norm2)
  cos_theta <- pmin(pmax(cos_theta, -1), 1) # Numerical clamping to avoid acos domain errors
  angle <- acos(cos_theta)
  return(angle)
}
```

---

## References

- Fisher, N. I. *Statistical Analysis of Circular Data* (1993)  
- Hyndman & Fan (1996) *The American Statistician* 50(4): 361–365.  
- Tensor Omics internal notes on RAP projections and shift vectors.

**Assignees:** @schroederaaron, @VivianBass

*Migrated from GitLab: https://gitlab.rlp.net/a.hallab/tensor-omics/-/work_items/87*

qwerdenkerXD:
## Summary
Adds directional/angle-based outlier detection, ported onto current `main`:

- `tox_get_outliers_by_angle` — spherical-mean-direction outlier detection for gene expression vectors (unit-vector angle to family reference direction, scaled by angular dispersion, percentile threshold)
- `tox_get_outliers_by_angle_rap` — the circular-statistics counterpart for RAP angles (circular mean, angular deviation, same scaling/threshold pipeline)
- Python and R/Rcpp bindings for both modules

This branch had diverged from `main` before the `src/` reorg (flat → `src/f42/`, `src/tox/`), the loess integration, and the parallelization rework, so a straight 3-way merge wasn't practical — most generic `test/`/`src/` files had been independently rewritten on both sides. Instead, the angle-outlier feature itself was replayed onto current `main`:

- New modules placed at their current `src/tox/` locations; new test files use `.F90` per main's current convention, registered in `test/run_tests.F90`
- Ported the handful of shared-module dependencies main didn't have yet: `wrap_angle` (`f42_utils.F90`), `STAT_NO_STABLE_DIRECTION`/`STAT_NO_ANGULAR_VARIATION` (`tox_errors.F90`), `tox_normalize_vectors_unit_length(_helper/_c)` (`tox_normalization.F90`)

While porting, found and fixed two pre-existing bugs in the original branch's Python bindings: `tox_angle_outliers`'s ctypes call passed a trailing `status` argument the Fortran routine never declared, and `tox_angle_outliers_rap_expert`'s ctypes signature (`percentile`/`valid_scores`/`perm`/`count_valid`) didn't match the actual 5-argument Fortran routine at all (`z_scores, threshold, is_outlier, n_genes, ierr`). The latter was never exercised by the original test suite, which is presumably why it went unnoticed. Both are now fixed to match the real Fortran signatures, mirroring the already-correct sibling wrappers.

## Test plan
- [x] Full Fortran suite (`test_runner.sh --compiler=gfortran`) passes, including the two new suites `get_outliers_by_angle` and `get_outliers_by_angle_rap`
- [x] Full Python test suite passes, including the two new/ported test files
- [x] Full R test suite passes, including the new `mod_test_get_outliers_by_angle.R`

Closes #88


src/tox/tox_get_outliers_by_angle.F90
```diff
@@ -0,0 +1,929 @@
...
+        logical, dimension(n_genes), intent(out) :: is_outlier
+            !! Output boolean array indicating outlier status per gene
+
+        integer(int32) :: i_gene
+
+        ! Identify outliers
+        do concurrent(i_gene=1:n_genes) shared(z_scores, is_outlier, threshold)
+            is_outlier(i_gene) = z_scores(i_gene) >= threshold
+        end do
+    end subroutine tox_angle_outliers_helper
```
schroederaaron, line 434: Same issue as in the non-angle outlier detection. Using a fixed percentile requires the user to know the percentage of outliers. If we set our outlier threshold to 0.95, we will _always_ get 5% outliers. No matter if in reality it is only 2% or maybe even 20%. A large angular deviation does not necessarly mean a biologically relevant shift. Some genes may vary more naturally, where a change of 20° may not be relevant, while on the other hand very stable genes may be biologically meaningfull with a change of 10°. Therefore, this outlier detection only returns the largest deviations, ignoring biological differences in gene expression. What do you think @asishallab ? Should we keep this for now like it is? I think we could potentially re-use the noise model (when it works properly) or a slightly modified version. Since we could theoretically again pool and check "normal" deviations of similar genes and then test if the observed deviations exceeds the expected deviation.

schroederaaron (changes requested):
Before i review everything the open thread should be solved

src/tox/tox_get_outliers_by_angle.F90
qwerdenkerXD: Well, that's why the threshold needs to be passed. The way on how the threshold is being calculated is the user's choice. It can be a percentile, but doesn't need to be.

Anyway, if we ever want a specific default method to determine the threshold, we can still update the module. I don't think it's worth to keep this PR open until we developed a good method for that

schroederaaron: I agree. Should we discuss it on Tuesday and keep the PR as it is until then?

qwerdenkerXD: ok, I'll mention in help channel as well

qwerdenkerXD: as discussed, this won't be neither a part of this branch nor a blocking condition for merging

rcpp/tensoromics_functions.cpp
```diff
@@ -4545,6 +4702,550 @@ List tox_detect_outliers_rcpp(NumericVector distances, IntegerVector gene_to_fam
...
+        double min_angular_dispersion,
+        double max_angular_dispersion) {
+
+    int n_samples = expression_vectors.nrow();
+    int n_genes   = expression_vectors.ncol();
+
+    // Compute n_families safely
+    int n_families = 0;
+    if (gene_to_fam.size() > 0)
+        n_families = Rcpp::max(gene_to_fam);
```
schroederaaron: Not sure on this, but wouldn't it be better if you count how many unique numbers you have that are non zero? What if in a previous filtering step every gene from a specific family was excluded? That way you have less families than the highest number, wouldn't you?

rcpp/tensoromics_functions.cpp
```diff
@@ -4545,6 +4702,550 @@ List tox_detect_outliers_rcpp(NumericVector distances, IntegerVector gene_to_fam
...
+        double min_angular_dispersion,
+        double max_angular_dispersion) {
+
+    int n_samples = unit_vectors.nrow();
+    int n_genes   = unit_vectors.ncol();
+
+    // Compute n_families safely
+    int n_families = 0;
+    if (gene_to_fam.size() > 0)
+        n_families = Rcpp::max(gene_to_fam);
```
schroederaaron: I think in the expert version `n_families` is something we can require from the user so we do not need to iterate over the full array

rcpp/tensoromics_functions.cpp
```diff
@@ -4545,6 +4702,550 @@ List tox_detect_outliers_rcpp(NumericVector distances, IntegerVector gene_to_fam
...
+        Rcpp::IntegerVector gene_to_fam,
+        double min_family_dispersion,
+        double max_family_dispersion) {
+
+    int n_genes = rap_angles.size();
+
+    // Compute n_families safely
+    int n_families = 0;
+    if (gene_to_fam.size() > 0)
+        n_families = Rcpp::max(gene_to_fam);
```
schroederaaron: Same as earlier, in expert functions `n_families` can be a required input. It is a very common variable so most users will have it saved anyways

rcpp/tensoromics_functions.R
schroederaaron: Why are you validating all inputs here if they are also validated on fortrans side?

src/tox/tox_get_outliers_by_angle.F90
```diff
@@ -0,0 +1,930 @@
...
+              tox_compute_angles_to_direction, tox_z_scores_by_dispersion, &
+              tox_angle_outliers, tox_detect_angle_outliers_pipeline
+    public :: tox_compute_family_direction_alloc, tox_angle_outliers_alloc
+
+#define CM_MIN_ANGULAR_DISPERSION_DEFAULT 1.0e-02_real64
+#define CM_MAX_ANGULAR_DISPERSION_DEFAULT sqrt(-2.0_real64 * log(0.5_real64))
+
+#define CM_ANGLES_SENTINEL -1.0_real64
+#define CM_ANGULAR_DISPERSION_SENTINEL -1.0_real64
+#define CM_Z_SCORES_SENTINEL -1.0_real64
```
schroederaaron: defined twice, if one is changed but not the other it may cause confusion. Should be centralized somewhere

src/tox/tox_get_outliers_by_angle.F90
```diff
@@ -0,0 +1,930 @@
...
+                                                 family_directions, angular_dispersions, &
+                          n_samples, n_genes, n_families, status, tmp_member_counts, min_angular_dispersion, max_angular_dispersion)
+    end subroutine tox_compute_family_direction_alloc
+
+    !> Compute angles between each gene and its family reference direction
+    pure subroutine tox_compute_angles_to_direction(unit_vectors, family_directions, &
+                                                    gene_to_fam, angles, &
+                                                    n_samples, n_genes, n_families, ierr)
+        integer(int32), intent(in) :: n_samples
+            !! Number of samples (dimension of vectors)
```
schroederaaron: dimension of vectors comment should be removed

src/tox/tox_get_outliers_by_angle.F90
```diff
@@ -0,0 +1,930 @@
...
+
+    end subroutine tox_compute_family_direction_helper
+
+    !> Compute family reference direction (spherical mean) and angular dispersion.
+    !| Some directions might result in sentinel values. In such case the respective family is being marked by the `status(i_family)` set to the respective status code.
+    pure subroutine tox_compute_family_direction_alloc(unit_vectors, gene_to_fam, &
+                                                       family_directions, angular_dispersions, &
+                                       n_samples, n_genes, n_families, ierr, status, min_angular_dispersion, max_angular_dispersion)
+        integer(int32), intent(in) :: n_samples
+            !! Number of samples (dimension of vectors)
```
schroederaaron: dimension of vectors part should be removed

src/tox/tox_get_outliers_by_angle.F90
```diff
@@ -0,0 +1,930 @@
...
+                          n_samples, n_genes, n_families, status, tmp_member_counts, min_angular_dispersion, max_angular_dispersion)
+    end subroutine tox_compute_family_direction
+
+    !> (no input validation) Compute family reference direction (spherical mean) and angular dispersion.
+    !| Some directions might result in sentinel values. In such case the respective family is being marked by the `status(i_family)` set to the respective status code.
+    pure subroutine tox_compute_family_direction_helper(unit_vectors, gene_to_fam, &
+                                                        family_directions, angular_dispersions, &
+                          n_samples, n_genes, n_families, status, tmp_member_counts, min_angular_dispersion, max_angular_dispersion)
+        integer(int32), intent(in) :: n_samples
+            !! Number of samples (dimension of vectors)
```
schroederaaron: dimension of vectors part should be removed

src/tox/tox_get_outliers_by_angle.F90
```diff
@@ -0,0 +1,930 @@
...
+
+contains
+
+    !> Compute family reference direction (spherical mean) and angular dispersion.
+    !| Some directions might result in sentinel values. In such case the respective family is being marked by the `status(i_family)` set to the respective status code.
+    pure subroutine tox_compute_family_direction(unit_vectors, gene_to_fam, &
+                                                 family_directions, angular_dispersions, &
+                    n_samples, n_genes, n_families, ierr, status, tmp_member_counts, min_angular_dispersion, max_angular_dispersion)
+        integer(int32), intent(in) :: n_samples
+            !! Number of samples (dimension of vectors)
```
schroederaaron: dimension of vectors part should be removed

src/tox/tox_get_outliers_by_angle_rap.F90
```diff
@@ -0,0 +1,836 @@
...
+
+        call set_ok(status)
+
+        family_mean_angles = 0.0_real64
+        family_dispersions = 0.0_real64
+        tmp_member_counts = 0_int32
+
+        ! Reusing arrays for precalculations
+        sum_sin_ptr => family_mean_angles
+        sum_cos_ptr => family_dispersions
```
schroederaaron: what exactly are you using the pointers for? As far as i can see it here, you are directly pointing to the full array, so, why don't just write in the array directly? `sum_sin_ptr` is just an alias for `family_mean_angles`, both have the exact same memory. Writing in `sum_sin_ptr` is like writing in the array directly. Also, using multiple variables pointing to the same memory is possibly not really safe to use in do concurrent, as you never know what compiler under what optimization reorders things in what way. If i did not miss anything that justifies the pointer usage, i highly recommend replacing them by direct array access.

schroederaaron (changes requested):

schroederaaron (changes requested):
Minor things, but i think the pointers should be removed

qwerdenkerXD:
Replaced by #220, which replays this feature onto current main on the code generator. The generator now writes the bindings, so the hand-written ctypes/Rcpp/_c code here is not carried over.

All review points from this PR are addressed there:
- n_families is a caller argument;
- constants are defined once;
- no R re-validation;
- no pointers;
- no "dimension of vectors" fragments.

The branch `87-outlier-detection-by-angle` is left as it is.

PR #163, opened 2026-07-31
Change calc_percentile argument range from [0,100] to [0,1] (#149)

qwerdenkerXD:
Closes #149.

## What

Every percentile/quantile argument in the framework now takes a probability/quantile in the statistical interval **[0,1]** instead of a percentage in [0,100]. There are **no `/100` conversions left anywhere** in the codebase (verified by audit).

Three distinct public arguments were converted, plus the internal `calc_percentile` family:

1. **`calc_percentile` family** (`f42_utils`) — the routine named in the issue
2. **`residual_range_quantile`** (`determine_shared_residual_range` in the JSD module)
3. **`percentile`** of `identify_outliers` / `detect_outliers`

## Commits (one logical case per commit, per layer)

| Commit | Scope |
|--------|-------|
| Change calc_percentile range | `calc_percentile` / `_alloc` / `_helper` / `_rank`: validation `100→1`, drop `/100` in the rank formula, docs; + 4 direct callers (`tox_get_outliers` `5.0→0.05`/`1.0→0.01`, `preprocessing` drop `·100`, `jsd` hand-off) |
| Add calc_percentile unit tests | new `test/mod_test_percentile.F90` (8 tests): min/median/max, interpolation, inclusive [0,1] boundaries, `_alloc` on unsorted input, and a **regression test that old percent-scale values (50.0, 95.0) are now rejected** |
| residual_range_quantile → [0,1] | Fortran (validation `100→1`, default `95.0→0.95`, remove the interim `/100` hand-off), then Python wrapper+test, then R wrapper+test |
| identify/detect_outliers → [0,1] | Fortran (defaults `95.0→0.95`, `perc_pos = n·percentile`, C-binding docs), then Python wrapper+test, then R wrapper+test |

## Verification that percentile-dependent implementations are ported correctly

All numeric results are **unchanged** — a fraction `p` reproduces the previous `100·p` percentage exactly. This is backed by existing value-pinning tests that pass with the converted inputs:
- `test_determine_shared_residual_range`: R = 10.65 (0.95), 4.0 (q=0.5), 11.0, 3.95; out-of-range now probes `above(1.0)`.
- `test_pool_means_alloc_basic`: x_star = 3.25 / 5.5 / 7.75.
- `get_outliers` suite: thresholds at 0.75 / 1.0 / 0.5 / 0.9 nearest-rank positions.

## Testing

`bash run_all_tests.sh` — the complete framework is green:
- Fortran: all 33 suites pass with gfortran (incl. the new `percentile` suite).
- Python: all 24 test files `success`.
- R: all 25 test files `success`.

Ford-generated `doc/` and `snippets/` regenerate from the updated source comments; not hand-edited here.

### DIFF
--- a/python/test/mod_test_data_integration_jsd.py
+++ b/python/test/mod_test_data_integration_jsd.py
@@ -42,13 +42,13 @@ def test_tox_determine_shared_residual_range():
     S2[:, 0, 1] = [5,  7,  9]
     S2[:, 1, 1] = [0,  1,  2]
 
-    R = tox_determine_shared_residual_range(S1, S2, 95.0)
+    R = tox_determine_shared_residual_range(S1, S2, 0.95)
     assert abs(R - 10.65) < TOL, f"Test 1 failed: expected 10.65, got {R}"
 
     # ============================================================
-    # Test 2 — Custom quantile (50%)
+    # Test 2 — Custom quantile (median, q=0.5)
     # ============================================================
-    R = tox_determine_shared_residual_range(S1, S2, 50.0)
+    R = tox_determine_shared_residual_range(S1, S2, 0.5)
     assert abs(R - 4.0) < TOL, f"Test 2 failed: expected 4.0, got {R}"
 
     # ============================================================
@@ -57,9 +57,9 @@ def test_tox_determine_shared_residual_range():
     assert_error(lambda: tox_determine_shared_residual_range(S1, S2, -1.0), "Test 3 failed: expected ERR_INVALID_INPUT")
 
     # ============================================================
-    # Test 4 — Quantile > 100 → error
+    # Test 4 — Quantile > 1 → error
     # ============================================================
-    assert_error(lambda: tox_determine_shared_residual_range(S1, S2, 150.0), "Test 4 failed: expected ERR_INVALID_INPUT")
+    assert_error(lambda: tox_determine_shared_residual_range(S1, S2, 1.5), "Test 4 failed: expected ERR_INVALID_INPUT")
 
     # ============================================================
     # Test 5 — NaNs must be ignored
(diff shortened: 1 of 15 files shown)


python/test/mod_test_data_integration_jsd.py
```diff
@@ -57,9 +57,9 @@ def test_tox_determine_shared_residual_range():
     assert_error(lambda: tox_determine_shared_residual_range(S1, S2, -1.0), "Test 3 failed: expected ERR_INVALID_INPUT")
 
     # ============================================================
-    # Test 4 — Quantile > 100 → error
+    # Test 4 — Quantile > 1 → error
     # ============================================================
-    assert_error(lambda: tox_determine_shared_residual_range(S1, S2, 150.0), "Test 4 failed: expected ERR_INVALID_INPUT")
+    assert_error(lambda: tox_determine_shared_residual_range(S1, S2, 1.5), "Test 4 failed: expected ERR_INVALID_INPUT")
```
schroederaaron, line 62: Just a general thought, not necessarily relevant for this PR. Wouldn't it make sense to also let the assert_error() function accept a certain error code, e.g.

`assert_error(lambda: function(something), 201, "Test failed, expected ....")` or even with textual form of the error code? That way we wouldn't only assert that some error is raised, but that the correct error is raised, which may be relevant for subroutines with multiple possible error codes

schroederaaron (approved):

python/test/mod_test_data_integration_jsd.py
qwerdenkerXD: I'll take that into the generator. We have custom error types there knowing their error code.

qwerdenkerXD: [See here](https://github.com/asishallab-group/Tensor-Omics/pull/153#discussion_r3719891717)

schroederaaron: perfect. sounds good.

PR #164, opened 2026-07-31
Audit shared-ierr in do concurrent loops; parallelize log-transform loops (#158)

qwerdenkerXD:
Closes #158.

## Audit result

Reviewed **every** `do concurrent` loop that touches an `ierr`/status code.

**No `do concurrent` loop currently writes a shared scalar `ierr`.** The race
flagged in PR #135 (`tox_errors.F90`) was the only one and was already fixed
there. Every `do concurrent` loop that calls a subroutine (euclidean distance,
gene centroids, trajectory velocity, JSD percentile, …) already writes into a
distinct per-iteration array slot via a **non-validating** helper — the
established safe pattern.

## What this PR changes

Three loops had been kept **sequential** *solely* because they folded a shared
`ierr` written on the error path of `logx` (which validates its own argument).
That was the only blocker to parallelizing them. Following the "non-validating
helper" convention used everywhere else:

- Split the pure math out of `logx` into **`logx_helper`** (no validation, no
  `ierr`); `logx` keeps its validating contract by delegating to it.
- **`log2_transformation_inplace_helper`** — validate `x > -1` up front (bulk,
  sequential — the accepted place for shared-`ierr` validation), then run the
  `log2(x+1)` transform as a `do concurrent` calling `logx_helper`.
- **`compute_family_scaling`** (two loops) — validate the log2 arguments up
  front, exactly as `logx` would (`> 0`, reject NaN/Inf), then call
  `logx_helper` under `do concurrent`. The callee does **not** assume the
  caller's `distances` are well-formed: a NaN/Inf distance propagates into a
  family mean and is now caught (previously `compute_family_scaling` validated
  only family *indices*, never distance *values*). The raw means validated for
  the first loop are the same values the second loop reads via `tmp_means_aux`,
  so one up-front check provably covers both loops.

## Shared-`ierr` cases intentionally left sequential (none are unresolved races)

Every case where a shared `ierr` was the *sole* blocker to parallelism is now
resolved. The remaining shared-`ierr` writes are all sequential for reasons
that removing `ierr` would not change:

- **Validation routines** (`tox_errors`, `tox_data_validation`): the loop's
  only job is to accumulate the error code — this is the accepted convention.
- **Sequential by construction**: file I/O (`tox_data_tools`,
  `tox_data_archive`), hashset/linked-list traversal (`f42_xxh3_hashmap`),
  subset generation/recursion (`tox_paralog_analysis`), and loops reusing a
  single shared scratch buffer (`tox_conversions` string buffer,
  `tox_trajectory_contribution_analysis` `tmp_*` buffers). `ierr` is incidental.
- `tox_trajectory_normalization` already uses the per-element `status(...)`
  **array** pattern the issue cites as the reference.

## Tests

- `test_log2_below_minus_one_errors` — covers the log2 up-front validation
  (`x = -1` and `x < -1` → `ERR_INVALID_INPUT`).
- `test_family_scaling_non_finite_distance` — a +Inf distance across two valid
  families now errors instead of silently reaching `log2`.
- Existing `log2_transformation`, `get_outliers`, and `normalization_pipeline`
  suites (which exercise the converted loops) all pass; the exact-value log2
  tests confirm the `do concurrent` path is numerically identical.

### DIFF
--- a/src/f42/f42_utils.F90
+++ b/src/f42/f42_utils.F90
@@ -257,12 +257,30 @@ contains
 
         if (is_err(ierr)) return
 
+        call logx_helper(val, base, exponent)
+    end subroutine logx
+
+    !> AUTHOR_FRANZ_ERIC_SILL
+    !| (no input validation) Compute logarithm for any base: `log_base(val) = log(val)/log(base)`.
+    !| This is the non-validating core of [[f42_utils(module):logx(subroutine)]], split out so it can
+    !| be called from inside `do concurrent` loops without a shared `ierr` (which would be a data
+    !| race). The caller is responsible for guaranteeing valid inputs beforehand -- `val > 0`,
+    !| `base > 0` and `base /= 1` -- e.g. via an up-front [[tox_errors(module):validate_in_range_real(subroutine)]]
+    !| pass; violating them yields a NaN/Inf result instead of an error code.
+    pure subroutine logx_helper(val, base, exponent)
+        real(real64), intent(in) :: val
+            !! Value (`x` in \( b^y = x \)), must be `> 0`
+        real(real64), intent(in) :: base
+            !! Base (`b` in \( b^y = x \)), must be `> 0` and `/= 1`
+        real(real64), intent(out) :: exponent
+            !! Exponent (`y` in \( b^y = x \))
+
         if (base == 2.0_real64) then
             exponent = log(val)/LOG_2
         else
             exponent = log(val)/log(base)
         end if
-    end subroutine logx
+    end subroutine logx_helper
 
     !> AUTHOR_FRANZ_ERIC_SILL
     !| Initialize Fortran's random number generator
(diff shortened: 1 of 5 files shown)


src/f42/f42_utils.F90
```diff
@@ -257,12 +257,30 @@ pure subroutine logx(val, base, exponent, ierr)
...
         if (is_err(ierr)) return
 
+        call logx_helper(val, base, exponent)
+    end subroutine logx
+
+    !> AUTHOR_FRANZ_ERIC_SILL
+    !| (no input validation) Compute logarithm for any base: `log_base(val) = log(val)/log(base)`.
+    !| This is the non-validating core of [[f42_utils(module):logx(subroutine)]], split out so it can
+    !| be called from inside `do concurrent` loops without a shared `ierr` (which would be a data
+    !| race). The caller is responsible for guaranteeing valid inputs beforehand -- `val > 0`,
```
schroederaaron: Not sure if this explanation is really needed. Maybe shorten it a bit e.g. only the section that it is the non-validating core. Keep the "can be called from inside do concurrent" only if we have that statement everywhere. I think it makes more sense to state if a subroutine can not be called, rather than stating if one can be called.

schroederaaron (approved):
Check the comments if all information is actually necessary. Some of them seem a bit excessive to me

PR #176, opened 2026-08-21
Median quickselect k d tree

asishallab:
Implemented more efficient K-D-Tree

    - The new version uses quickselect median based dimension splitting
    - It thus can be constructed in O(log(n) * n) instead of previously O(log^2(n) * n).

The branch was created from 131-codegen-new and cherry-picked K-D-Tree related methods from 

* shape-truthful-clustering
* shape-truthful-clustering-tangent-space-variant
* shape-truthful-clustering-updated
* smoothing

Should speed up almost everything we do in Tensor Omics.


qwerdenkerXD:
@asishallab #153 needs to be merged first, just that you know

PR #177, opened 2026-08-23
Make kd-tree impl more efficient by using quickselect

qwerdenkerXD:
Requires #131 to be merged.

This is a re-opening of #176


qwerdenkerXD:
@asishallab Please remove garbage from the branch. The included code generator brings around 90k additions, so there are 210k lines being added by this PR (300k in total), which is not reasonable for just making the kd tree implementation more efficient.

Issue #181, opened 2026-09-03
Recipes deferred from the manual, and what blocks each


qwerdenkerXD:
Running list of recipes that belong in the manual (#142, spec in #147) but could not be written yet, each with what blocks it. A recipe is only useful if it is correct, so where the library's behaviour or its naming is unsettled the recipe waits for a decision rather than guessing at one.

Add to this list as further recipes hit the same wall.

---

## 1. Tissue Versatility

**What it would document:** `compute_tissue_versatility` — how uniformly a gene is expressed across the selected axes, from the angle between its expression vector and the space diagonal.

**The formula is settled and the code matches it.** #39 specifies

$$\text{TV} = \frac{1-\cos\varphi}{1-\frac{1}{\sqrt{n}}}$$

and that is exactly what `compute_tissue_versatility_impl` computes. Verified:

| input (4 axes) | TV | angle |
|---|---|---|
| uniform `(1,1,1,1)` | 0 | 0° |
| two axes `(1,1,0,0)` | 0.586 | 45° |
| single axis `(1,0,0,0)` | 1 | 60° |

### Blocker A — the name runs opposite to the value

TV rises as a gene becomes more tissue-**specific**: `0` = uniform, `1` = one tissue only. So the quantity called *versatility* is a specificity score, and #39 contradicts itself on the consequence:

> "$0$: perfectly uniform expression. $1$: expression in only one tissue."
>
> "**Tissue Specificity** is defined as the complement of tissue versatility: $\text{Specificity} = 1 - \text{TV}$. Values close to 1 → gene is highly tissue-specific."

By those numbers a single-tissue gene has TV = 1 and therefore Specificity = 0 — the opposite of what the second quote claims. A reader who takes `1 - TV` as specificity, as #39 instructs, inverts their result.

The code has inherited the muddle. Its module doc says a gene confined to one tissue "points along that axis and **scores minimally**", while it scores 1, the maximum; the argument doc immediately below is correct ("0 means uniform expression and 1 means expression in only one axis"). The two disagree inside one file.

**Needed to unblock:** a decision on which of these is intended, since it changes what the recipe tells people to compute.

- keep the values, fix the vocabulary — the score is specificity, and versatility is its complement (documentation-only, no API change); or
- keep the name, invert the formula so TV = 1 means uniform (breaks the published Python/R/C API and every caller); or
- rename the procedure to `compute_tissue_specificity` (also an API break).

[...]

Issue #180, opened 2026-09-03
detect_outliers returns NaN in loess_x/loess_y, but TOX does not produce NaN


qwerdenkerXD:
`compute_family_scaling` pads `loess_x`/`loess_y` with NaN, and `detect_outliers` returns it to the caller.

Tensor Omics does not produce NaN as an output value — NaN is accepted on input (that is what `DM_ALLOW_NAN` marks), but no routine should write one into a result. This is one place that does, and the NaN reaches the public Python, R and C APIs.

## Where

`compute_family_scaling_impl` fits its LOESS over the `n_valid` families that have more than one member. When that is fewer than `n_families`, it pads the unused tail of its two reference arrays:

https://github.com/asishallab-group/Tensor-Omics/blob/147-manual-recipes/src/tox/tox_get_outliers_impl.F90#L389-L391

```fortran
if (n_valid < n_families) then
    loess_x(n_valid + 1:) = M_NAN
    loess_y(n_valid + 1:) = M_NAN
    indices_used(n_valid + 1:n_families) = 0_int32
end if
```

`loess_x` and `loess_y` are `intent(out)` and are returned by both `compute_family_scaling` and `detect_outliers`, so this is not internal scratch.

## Reproducing

Any dataset with a family that has a single member — ordinary after filtering — leaves slots unused:

```python
import numpy as np
from tensor_omics import group_centroid_all, distance_to_centroid, detect_outliers

rng = np.random.default_rng(4)
n_families = 40
sizes = [1, 1, 1] + [6] * (n_families - 3)   # three one-member families
gene_to_fam = np.asarray(np.repeat(np.arange(1, n_families + 1), sizes), dtype=np.int32)
expr = np.asfortranarray(rng.lognormal(0, .5, size=(4, len(gene_to_fam))))

centroids = group_centroid_all(expr, gene_to_fam, n_families)
distances = distance_to_centroid(expr, centroids, gene_to_fam)
result = detect_outliers(n_families, distances, gene_to_fam)

for key in ("loess_x", "loess_y"):
    print(key, int(np.isnan(result[key]).sum()), "NaN of", result[key].size)
```

```
loess_x 4 NaN of 40
loess_y 4 NaN of 40
```

## Suggested direction

[...]

Issue #142, opened 2026-05-28
Complete our Manual


asishallab:
# Complete our Tensor Omics Manual

The Tensor Omics Manual is not complete. It lacks the whole section on R, if I'm not mistaken. And some analyses are not yet in it, or are they?

Can you please check and report on this

VivianBass:
You can check out the manual here: [tox_manual](https://github.com/asishallab-group/Tensor-Omics/blob/general_documentation/misc/tox_manual.pdf)

It's not that the whole R section is missing, but it is outdated regarding the latest modules we've developed. Franz is currently working on a code generator inside the workflow to build the manual automatically, so we won't have to do it manually anymore.

Issue #123, opened 2026-02-11
Optional quantile normalization in pipeline


VivianBass:
Make quantile normalization optional and by default switched off.

Update fortran code, fortran tests, rcpp, r, python, and tests for r and python.

**Assignees:** @VivianBass

*Migrated from GitLab: https://gitlab.rlp.net/a.hallab/tensor-omics/-/work_items/122*

VivianBass:
This has been done in loess_integration branch

PR #182, opened 2026-09-03
Fill the recipe manual: eight recipes, with the library fixes writing them turned up

qwerdenkerXD:
Fills the recipe manual (#142) with real content, against the spec in #147. Eight recipes, 33 pages, each with a runnable example script under `python/examples/`.

**Based on `131-codegen-new`, not `main`, on purpose.** The recipes are written against the generated API that branch introduces, and this cannot land before #153 does. GitHub has no merge-dependency setting, so the base branch is the dependency: this PR is only mergeable into `131-codegen-new`, and GitHub will retarget it automatically once #153 merges. It also keeps the diff to this work rather than replaying the whole generator branch.

## The recipes

| | |
|---|---|
| **Analyses** | Detecting Divergent Genes (outliers + shift vectors) · Trajectory-Based Signed Contribution Analysis · SND detection |
| **Shared** | Creating a TOX Dataset · Normalization · Gene Family Detection · Gene Family Centroids · Clustering |

Both files that shipped as illustrative scaffolding — `snd_detection.tex` and `gene_family_detection.tex` — are now real recipes. Nothing illustrative is left in the manual.

Every number quoted in a recipe comes from a run, not from reading the code. Where a recipe's claim could be checked against an external reference it was: the linkage heights are compared against `scipy.cluster.hierarchy.linkage`, and OrthoFinder's output formats are taken from its own documentation.

## Method/implementation mismatches, resolved from closed issues

The methods document is the stale side more often than the code, and in each case a closed issue already settled it:

- **RDI.** The chapter defines `d / max‖fᵢ − o‖`; #27 chose the per-family standard deviation *and* asked for the smoothed-over-all-families version that is implemented, with #67 covering that same fit.
- **Normalization.** The chapter's three-regime "Scheme S" was rejected by #67; quantile normalization was made optional by #123.
- **Centroids, SND, TCA, clustering.** #8, #64, #99/#104 and #87 all match the code, except where noted below.

Two deviations are documented in the recipes rather than filed: the chapter's per-family median angle for the SND prefilters is not expressible (one global scalar), and the chapter's `+ε` in the trajectory normalization would stop the maximum reaching exactly 1, which the implementation's approach does not.

## Library fixes found while writing

Each was reproduced before being changed, and the full suite (57 modules) is green at every commit that touches `src/`.

- **`307dfd66`** — `compute_shift_vector_field` read `family_centroids(:, 0)` for any gene carrying the documented unassigned sentinel. Silent garbage in a release build; under `--diagnostics`, `Index '0' … below lower bound of 1`.
- **`b46ee4c8`** — outlier detection rejected that same sentinel, so a dataset holding genes outside the families under study could not be analysed at all. Accepting it required fixing the threshold too: the clamped zeros would have moved the 95th percentile to roughly the 58th percentile of the assigned genes. Also, 3–4 families made netlib's assertion `stop 1` and kill the host process.
- **`6c453a90`** — `save_tox_data` silently dropped an array whose `*_file` name was omitted, while documenting it as "will be saved if provided".
- **`991a00ef`** — `cluster_factor_trajectories_k_means` clustered `(sample, timepoint)` states rather than trajectories, contrary to #87, and was bit-identical to `k_means_clustering` on a hand-reshaped array.

Plus three documentation defects corrected at source, so all four language bindings carry them: `calc_fchange`'s input contract, `distance_to_centroid`'s sentinel note (attached to the wrong argument), and `tox_get_outliers`'s module doc.

## Still open

- **#181** collects the recipes that could not be written and what blocks each: tissue versatility (its name runs opposite to its value), RAP projection (three readings of a "field" across two modules — code deliberately left unchanged), JSD, empirical noise significance, space interpolation, batch effects.
- **#180** — `detect_outliers` returns NaN in `loess_x`/`loess_y`, which the no-NaN rule forbids.
- A comment on #153 about the Python emitter dropping procedure doc bodies that the roxygen emitter keeps.

No reviewers assigned yet.

https://claude.ai/code/session_01VfUFSLm6ELKrXdU2MqYSmn


qwerdenkerXD:
Depends on #153. This PR targets `131-codegen-new` rather than `main`, which is the closest GitHub has to GitLab's merge-request dependencies: it is not mergeable into `main` at all until #153 lands, and GitHub retargets it automatically when that happens.

